PR #78, opened 2026-04-12
feat: Simplify fetching in frontend

alex289:
## **Type of change**

- [ ] 🐛 Bug fix
- [ ] 🚀 New feature

## **Description**

- Add swr wrapper
- Improve fetch wrapper

Now only one import is required for fetching and parameters like the useSWR fetcher or the base api url are already set


frontend/src/lib/api.ts
```diff
@@ -50,13 +52,26 @@ export async function getErrorMessage(res: Response): Promise<string> {
...
 	if (!res.ok) {
 		throw new Error(await getErrorMessage(res));
 	}
 	return (await res.json()) as JSON;
 }
+
+export async function fetcher<JSON = any>(
+	url: string,
+	method: "GET" | "POST" | "PUT" | "DELETE",
+	data?: any,
```
timokoessler: Define a type like `Record<string, any>` so at least it can not be a function or class?
Add a optional argument to provide additional headers?

timokoessler (commented):
Cool, just a small suggestion.

timokoessler (approved):

PR #79, opened 2026-04-12
feat: Agent UI

maximbigler:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

Add ui for agents overview

## **Additional context**

Screenshot 2026-04-12 at 18-18-05 Agents


frontend/src/lib/agents.ts
```diff
@@ -0,0 +1,39 @@
...
+export enum AgentStatus {
+	Online = 0,
+	Offline = 1,
+}
+
+export interface Agent {
+	id: string;
+	name: string;
+	ip: string;
+	url: string;
```
alex289: I don't think we have a url for an agent?

frontend/src/lib/agents.ts
```diff
@@ -0,0 +1,39 @@
...
+	status: AgentStatus;
+	cpuUsage?: number;
```
alex289: Do we really want to retrieve this metric? @timokoessler

frontend/src/components/cards/agents/data-cards.tsx
```diff
@@ -0,0 +1,183 @@
...
+import { Input } from "@/components/ui/input";
+import { AgentStatus, deleteAgent, type Agent } from "@/lib/agents";
+import { toSearchableText } from "@/lib/utils";
+import { toast } from "sonner";
+
+interface AgentDataCardsProps {
+	data: Agent[];
+}
+
+function renderStatusBadge(status: AgentStatus) {
```
alex289: This should have a component name

frontend/src/components/cards/agents/data-cards.tsx
```diff
@@ -0,0 +1,183 @@
...
+		</div>
+	);
+}
+
+function metricValue(value?: number | string) {
+	if (value === undefined || value === "") {
+		return "n/a";
+	}
+
+	return String(`${value} %`);
```
alex289: This also works right?

```suggestion
	return `${value} %`;
```

frontend/src/components/cards/agents/data-cards.tsx
```diff
@@ -0,0 +1,183 @@
...
+								</CardAction>
+
+								<CardTitle>{agent.name}</CardTitle>
+								<CardDescription>{agent.ip}</CardDescription>
+							</CardHeader>
+
+							<hr />
+
+							<CardContent className="space-y-3">
+								{renderStatusBadge(agent.status)}
```
alex289: This should be called as a component

frontend/src/components/dialogs/upsert-agent.tsx
```diff
@@ -0,0 +1,173 @@
...
+	asDropdownItem?: boolean;
+}) {
+	const isEditing = !!agent;
+	const [isSubmitting, setIsSubmitting] = useState(false);
+	const [open, setOpen] = useState(false);
+
+	const form = useForm({
+		defaultValues: {
+			name: agent?.name ?? "",
+			url: agent?.url ?? "",
```
alex289: Does an agent really need a url? @timokoessler

alex289:
Bildschirmfoto 2026-04-12 um 18 58 21

I think this icon at the top left would look cool, wouldn't it?

frontend/src/components/cards/agents/data-cards.tsx
```diff
@@ -0,0 +1,183 @@
...
+								<div className="grid grid-cols-2 gap-2 text-xs">
+									<div className="rounded-lg border bg-muted/50 p-2">
+										<p className="flex items-center gap-1 text-muted-foreground">
+											<Cpu className="h-3 w-3" />
+											CPU Usage
+										</p>
+										<p className="mt-1 font-medium">{metricValue(agent.cpuUsage)}</p>
+									</div>
+
+									<div className="rounded-lg border bg-muted/50 p-2">
```
alex289, line 148: Should we add the number of apps on this agent here?

alex289: Like we did with the repositories

frontend/src/components/cards/agents/data-cards.tsx
```diff
@@ -0,0 +1,183 @@
...
+				<div className="text-sm text-muted-foreground">
+					{filteredAgents.length} of {data.length} agents
+				</div>
+
+				<div className="relative w-full sm:max-w-md">
+					<Search className="pointer-events-none absolute left-3 top-1/2 h-4 w-4 -translate-y-1/2 text-muted-foreground" />
+					<Input
+						value={searchQuery}
+						onChange={(event) => setSearchQuery(event.target.value)}
+						placeholder="Search agents by name, URL or IP..."
```
alex289: Adjusting to the already existing search bars

```suggestion
						placeholder="Search agents..."
```

frontend/src/components/cards/agents/data-cards.tsx
```diff
@@ -0,0 +1,183 @@
...
+				status: agent.status === AgentStatus.Online ? "online" : "offline",
+			};
+
+			return toSearchableText(searchableAgent).includes(query);
+		});
+	}, [searchQuery, data]);
+
+	return (
+		<div className="space-y-4">
+			<div className="flex flex-col gap-3 sm:flex-row sm:items-center sm:justify-between">
```
alex289, line 77: Can we search bar and agent count be switched. Other search bars are always left

alex289 (changes requested):

frontend/src/components/dialogs/upsert-agent.tsx
timokoessler: Nope, it does not have any.

frontend/src/lib/agents.ts
timokoessler: Maybe later, but not sure if we want to store this in the db. I would keep this to the existing db schema for now.

frontend/src/components/cards/agents/data-cards.tsx
```diff
@@ -0,0 +1,183 @@
...
+				<div className="text-sm text-muted-foreground">
+					{filteredAgents.length} of {data.length} agents
```
alex289: Isn't this more confusing. In my opinion this would be helpful when having some kind of filter dropdowns or something but I wouldn't expect this to be applied by searching in the input. You can resolve if you think otherwise :D

frontend/src/components/dialogs/upsert-agent.tsx
```diff
@@ -0,0 +1,172 @@
...
+	asDropdownItem?: boolean;
+}) {
+	const isEditing = !!agent;
+	const [isSubmitting, setIsSubmitting] = useState(false);
+	const [open, setOpen] = useState(false);
+
+	const form = useForm({
+		defaultValues: {
+			name: agent?.name ?? "",
+			ip: agent?.ip ?? "",
```
alex289: Im not sure if we want to require setting the ip here. I thought we would infer it by getting it by the agent itself. @timokoessler ?

alex289: Ah lol we don't even have the ip in the db yet

maximbigler: Oh, yes, you're right. I got something mixed up. But how should the Hub know where to fetch the data?

alex289: The agent connects via websocket and this connection is kept alive so the hub can respond to that. Only the agent needs to know where to connect to

alex289:
I just yolod a PR for the agent endpoints with codex. I hope thats helpful
https://github.com/OrcaCD/orca-cd/pull/89

maximbigler:
Just a little update of the UI

Screenshot 2026-04-13 at 22-28-46 Agents

frontend/src/lib/agents.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export enum AgentStatus {
+	Online = 0,
+	Offline = 1,
+}
+
+export interface Agent {
+	id: string;
+	name: string;
+	ip: string;
```
alex289: I think we should remove ip and dockerVersion for now and readd it when we have it on the backend side

alex289: You can maybe comment it out so we have it already there if we add it

alex289 (approved):

PR #82, opened 2026-04-12
feat: Demo mode

alex289:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

Add a demo mode:

- Disable all POST routes except login and logout
- Seed db with example data
- Make db read only
- Add tests


backend/internal/hub/db/db.go
```diff
@@ -44,6 +46,33 @@ func Connect(logger zerolog.Logger, debug bool) error {
...
+
+		sqlDB, err := db.DB()
+		if err != nil {
+			return err
+		}
+		if err := sqlDB.Close(); err != nil {
+			return err
+		}
+
+		readOnlyDB, err := gorm.Open(sqlite.Open(dbPath+"?mode=ro"), &gorm.Config{
```
timokoessler: Can we define the config and pass it too both gorm.Open calls?

backend/internal/hub/db/demo_seed.go
```diff
@@ -0,0 +1,76 @@
...
+
+		user := models.User{
+			Base:                   models.Base{Id: demoSeedUserID},
+			Email:                  demoSeedUserEmail,
+			Name:                   demoSeedUserName,
+			Role:                   models.UserRoleAdmin,
+			PasswordChangeRequired: false,
+			PasswordHash:           &hashedPassword,
+		}
+		if err := tx.Clauses(clause.OnConflict{DoNothing: true}).Select("*").Create(&user).Error; err != nil {
```
timokoessler: Alternatively truncate DB on each start?

backend/internal/hub/db/demo_seed.go
```diff
@@ -0,0 +1,76 @@
...
+			return err
+		}
+
+		var userIDs []string
+		if err := tx.Model(&models.User{}).Where("email = ?", demoSeedUserEmail).Limit(1).Pluck("id", &userIDs).Error; err != nil {
```
timokoessler: Use new Gorm generics API, like we do in most places? https://gorm.io/docs/transactions.html#Generics-API

backend/internal/hub/server.go
```diff
@@ -67,16 +69,19 @@ func DefaultConfig() (Config, error) {
 		return Config{}, errors.New("invalid app secret: must be minimum 32 characters")
 	}
 
+	trueString := "true"
+
 	return Config{
-		Debug:            debug == "true",
+		Debug:            debug == trueString,
```
timokoessler: Maybe we can create a global helper function that e.g. resolves `true`, `yes`, `1`, `True` to true?

timokoessler (commented):
Cool PR!

backend/internal/hub/db/demo_seed.go
alex289: I added a check so if there is a user already we don't seed

timokoessler (approved):

PR #84, opened 2026-04-13
feat: Add translation system

timokoessler:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

No language picker added yet. To discuss whether we want to sync this setting or not? Also only translates login page to keep PR small.


frontend/messages/en.json
```diff
@@ -0,0 +1,46 @@
+{
+	"$schema": "https://inlang.com/schema/inlang-message-format",
```
alex289, line 2: Can translations be grouped here so we can have one object per page for example?

frontend/.oxlintrc.json
```diff
@@ -6,7 +6,6 @@
 	},
 	"rules": {
 		"no-alert": "error",
-		"no-console": "error",
```
alex289, line 9: Don't we want to try not using the console? Maybe better to ignore at the calling places

frontend/messages/en.json
timokoessler: Not supported :(

alex289 (approved):

PR #90, opened 2026-04-13
feat: Support gitea repositories

alex289:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

- Add gitea provider
- Enable gitea in ui
- Add tests

### DIFF
--- a/backend/internal/hub/repositories/gitea.go
+++ b/backend/internal/hub/repositories/gitea.go
@@ -1,0 +1,115 @@
+package repositories
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/OrcaCD/orca-cd/internal/hub/models"
+	"github.com/OrcaCD/orca-cd/internal/shared/httpclient"
+)
+
+type giteaProvider struct{}
+
+func init() {
+	Register(models.Gitea, giteaProvider{})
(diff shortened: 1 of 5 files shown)


backend/internal/hub/repositories/gitea.go
```diff
@@ -0,0 +1,117 @@
...
+	path := strings.TrimPrefix(u.Path, "/")
+	path = strings.TrimSuffix(path, ".git")
+
+	parts := strings.SplitN(path, "/", 3)
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+		return "", "", fmt.Errorf("URL must be in the format %s://{host}/{owner}/{repo}", httpsScheme)
+	}
+
+	owner, repo = parts[0], parts[1]
+	if !repoRe.MatchString(owner) {
```
timokoessler: Intended to use repo regex for both?

timokoessler: Maybe nicer to create a local regex and rename it in GitHub file so it's independent? (bad naming is my fault)

alex289: Gitlab was also using the regex. I moved the regex to the provider.go and if we need to customize it I would create a local regex in each provider

timokoessler (approved):

PR #93, opened 2026-04-14
sec: Fix prototype pollution on login page

timokoessler:


timokoessler:
Probably not a real vulnerability, but allows to crash rendering.

alex289 (approved):

PR #102, opened 2026-04-20
feat: Branch and file selector

alex289:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

- Add endpoints for getting repository branches and trees
- Show branch and file selector

## **Additional context**

- Disable goconst because it's too strict

### DIFF
--- a/backend/.golangci.yml
+++ b/backend/.golangci.yml
@@ -16,6 +16,5 @@ linters:
     - zerologlint
     - gocritic
     - bodyclose
-    - goconst
     - modernize
     - unparam
--- a/backend/internal/hub/repositories/gitea.go
+++ b/backend/internal/hub/repositories/gitea.go
@@ -2,10 +2,12 @@ package repositories
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
+	"sort"
 	"strings"
 
 	"github.com/OrcaCD/orca-cd/internal/hub/models"
@@ -13,6 +15,17 @@ import (
 )
 
 type giteaProvider struct{}
+
+type giteaBranch struct {
+	Name string `json:"name"`
+}
+
+type giteaTreeResponse struct {
+	Tree []struct {
+		Path string `json:"path"`
+		Type string `json:"type"`
+	} `json:"tree"`
+}
 
 const httpsScheme = "https"
 
@@ -83,15 +96,7 @@ func (giteaProvider) TestConnection(ctx context.Context, repo *models.Repository
 	if err != nil {
 		return fmt.Errorf("failed to build Gitea request: %w", err)
 	}
-
-	req.Header.Set("Accept", "application/json")
-
-	if repo.AuthMethod == models.AuthMethodToken && repo.AuthToken != nil {
-		token := strings.TrimSpace(repo.AuthToken.String())
-		if token != "" {
-			req.Header.Set("Authorization", "token "+token)
-		}
-	}
+	addGiteaHeaders(req, repo)
 
 	resp, err := httpclient.Default.Do(req)
 	if err != nil {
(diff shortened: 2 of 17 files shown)


backend/internal/hub/repositories/gitea.go
```diff
@@ -115,3 +120,177 @@ func (giteaProvider) TestConnection(ctx context.Context, repo *models.Repository
...
+		return nil, errors.New("repository is required")
+	}
+
+	owner, repoName, err := parseGiteaURL(repo.Url)
+	if err != nil {
+		return nil, fmt.Errorf("invalid repository URL: %w", err)
+	}
+
+	u, _ := url.Parse(repo.Url)
+	baseURL := fmt.Sprintf("%s://%s", u.Scheme, u.Host)
```
timokoessler: Does it make sense to add a helper method for this or return this in parseGiteaURL so that we don't parse the url twice? Might apply to Gitlab too tho.

backend/internal/hub/repositories/gitea.go
```diff
@@ -115,3 +120,177 @@ func (giteaProvider) TestConnection(ctx context.Context, repo *models.Repository
...
+
+	branches := make([]string, 0)
+	limit := 100
```
timokoessler: Add as const to provider.go and use for all providers?

backend/.golangci.yml
```diff
@@ -16,6 +16,5 @@ linters:
     - zerologlint
     - gocritic
     - bodyclose
-    - goconst
```
timokoessler, line 19: Just ignore for test files maybe? Or did it trigger on another file?

backend/internal/hub/repositories/gitlab.go
```diff
@@ -121,3 +127,219 @@ func (gitlabProvider) TestConnection(ctx context.Context, repo *models.Repositor
...
+	u, _ := url.Parse(repo.Url)
+	baseURL := fmt.Sprintf("%s://%s", u.Scheme, u.Host)
+	projectPath := url.PathEscape(namespace + "/" + project)
+
+	branches := make([]string, 0)
+	page := 1
+
+	for {
+		apiURL := fmt.Sprintf(
+			"%s/api/v4/projects/%s/repository/branches?per_page=100&page=%d",
```
timokoessler: No max like for Gitea?

backend/internal/hub/repositories/github.go
```diff
@@ -110,3 +116,139 @@ func (githubProvider) TestConnection(ctx context.Context, repo *models.Repositor
...
+	if repo == nil {
+		return nil, errors.New("repository is required")
+	}
+
+	owner, repoName, err := parseGitHubURL(repo.Url)
+	if err != nil {
+		return nil, fmt.Errorf("invalid repository URL: %w", err)
+	}
+
+	apiURL := fmt.Sprintf("%s/repos/%s/%s/branches?per_page=100", githubAPIBase, owner, repoName)
```
timokoessler: We only get 100 branches here (fine for me for now), but we fetch multiple pages for other providers?

backend/.golangci.yml
alex289: It triggered for a lot of things and I noticed it in some other prs too which was annoying. I think we can do this in review

frontend/src/components/dialogs/upsert-application.tsx
```diff
@@ -228,17 +396,43 @@ export default function UpsertApplicationDialog({
...
+													placeholder={
+														repositoryId ? "Select a branch" : "Select a repository first"
+													}
+												/>
+											</SelectTrigger>
+											<SelectContent position="item-aligned">
+												{isBranchesLoading ? (
+													<div className="p-2">Loading branches...</div>
+												) : (
+													branches?.map((branch) => (
```
timokoessler, line 429: Sort common branches (main, master, production, ...) before other branches? Maybe in backend?

frontend/src/components/dialogs/upsert-application.tsx
```diff
@@ -43,6 +45,144 @@ const applicationSchema = z.object({
...
+				node.type = "dir";
+			}
+		}
+	}
+
+	sortTreeNodes(root.children);
+	return root.children;
+}
+
+function TreeNodeList({
```
timokoessler, line 124: A bit dark on hover maybe?

Image

timokoessler, line 124: Not sure about it, but does it make sense to hide folders without yaml files and hide all non yaml / yml files? Better UX, but maybe looks nicer right now.

alex289: I think it's nice to show all files so in special cases the user can still overwrite the path e.g if a user uses other file extensions for whatever purposes

timokoessler: We should probably validate the path in the backend then? Check that it exists and looks like a Docker Compose yaml?

alex289: Does this make sense when the file could be deleted or changed at any time anyways?

timokoessler: Good question, maybe some minimal validation tho? At least that the path is not empty. For me it would also make sense to force yaml / yml extension.

backend/internal/hub/routes/applications.go
```diff
@@ -284,6 +298,44 @@ func toApplicationResponse(app *models.Application) applicationResponse {
...
+		}
+		return "", http.StatusInternalServerError, "internal server error"
+	}
+
+	provider, err := repositories.Get(repo.Provider)
+	if err != nil {
+		return "", http.StatusBadRequest, "unsupported provider"
+	}
+
+	entries, err := provider.ListTree(c.Request.Context(), &repo, branch)
```
timokoessler, line 326: Add todo to use "get single file" api in the future?

timokoessler (approved):

PR #106, opened 2026-04-21
feat: Fetch and store compose file

### LINKED ISSUE #28: feat: fetch compose file from git repository (public & private)
Define the architecture and protocol for agents to retrieve a `docker-compose.yml` (or equivalent) from a git repository on behalf of the hub. The agent runs on the target host and performs the actual fetch — the hub instructs it via a WebSocket connection using protobuf-encoded messages.

Only the compose file is fetched — no full clone. This minimizes disk usage, avoids credential sprawl (hub never touches repo credentials for private repos), and keeps the agent self-contained.

---

**Fetch Strategy (Agent-side)**

- For **public repos**: use the provider's raw file API — no auth needed
  - GitHub: `https://raw.githubusercontent.com/{owner}/{repo}/{ref}/{path}`
  - GitLab: `https://gitlab.com/{owner}/{repo}/-/raw/{ref}/{path}`
  - Gitea/Forgejo: `https://{host}/{owner}/{repo}/raw/branch/{ref}/{path}`
- For **private repos**: use a deploy token / fine-grained PAT scoped to read-only repository contents
  - Credentials are stored encrypted on the hub (existing `crypto` package) and sent to the agent as part of the protobuf message — never stored on the agent
  - HTTP `Authorization: Bearer <token>` or `Authorization: token <pat>` header
- Alternative for self-hosted / generic git: sparse checkout via `git clone --filter=blob:none --sparse --depth=1` then `git sparse-checkout set <file>` — heavier but universally compatible; use only as fallback

---

**Architecture Diagram**

```mermaid
sequenceDiagram
    participant User
    participant Hub
    participant Agent

    User->>Hub: Add application (repo URL, ref, file path, credentials)
    Hub->>Hub: Store credentials encrypted (AppSecret)

    Note over Hub,Agent: Persistent WebSocket connection<br/>(Agent dials Hub on startup)

    Hub->>Agent: WS frame — FetchComposeRequest (protobuf)<br/>{ repo_url, ref, file_path, access_token }
    Agent->>Agent: HTTP GET raw file URL<br/>(with Bearer token if private)
    Agent-->>Hub: WS frame — FetchComposeResponse (protobuf)<br/>{ success, content | error }

    Hub->>Hub: Parse & validate compose file
    Hub-->>User: Compose file ready / error feedback
```

---

**Considerations**

- `access_token` in the protobuf message is in-flight plaintext — **require TLS on the WebSocket (`wss://`)** in production; document this clearly
- The hub should sign or MAC each `FetchComposeRequest` with the shared agent secret so the agent can reject forged messages
- `request_id` (UUID v7) allows the hub to correlate async responses from multiple agents
- Credentials should never be logged on the agent side (`zerolog` field masking)
- File size limit should be enforced on the agent before returning content (e.g. reject files > 1 MB)
- Future: extend the same message pattern for `docker-compose up/down/pull` commands

alex289:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

Fixes #28 

- Fetch compose file
- Store compose file, commit hash and message in db


backend/internal/hub/repositories/gitea.go
```diff
@@ -296,6 +309,144 @@ func (giteaProvider) ListTree(ctx context.Context, repo *models.Repository, bran
 	}
 }
 
+func (giteaProvider) GetFileContent(ctx context.Context, repo *models.Repository, branch, path string) (string, error) {
```
timokoessler: Maybe not call it branch, but ref? We should fetch by commit hash on webhook push, to prevent edge cases where already another commit was pushed.

backend/internal/hub/repositories/provider.go
```diff
@@ -21,6 +21,15 @@ type Provider interface {
 	ListBranches(ctx context.Context, repo *models.Repository) ([]string, error)
 	// Lists repository tree entries for a branch.
 	ListTree(ctx context.Context, repo *models.Repository, branch string) ([]TreeEntry, error)
+	// Returns decoded file content for a repository file at the given branch.
+	GetFileContent(ctx context.Context, repo *models.Repository, branch, path string) (string, error)
```
timokoessler:
```suggestion
	GetFileContent(ctx context.Context, repo *models.Repository, ref string, path string) (string, error)
```

timokoessler (approved):

PR #115, opened 2026-04-21
feat: Add incoming webhook endpoint

timokoessler:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## Description

image

(Only tested GitHub for now)

### DIFF
--- a/backend/internal/hub/routes/webhooks.go
+++ b/backend/internal/hub/routes/webhooks.go
@@ -13,8 +13,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/OrcaCD/orca-cd/internal/hub/applications"
 	"github.com/OrcaCD/orca-cd/internal/hub/db"
 	"github.com/OrcaCD/orca-cd/internal/hub/models"
+	"github.com/OrcaCD/orca-cd/internal/hub/repositories"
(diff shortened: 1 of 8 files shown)


backend/internal/hub/routes/webhooks.go
```diff
@@ -0,0 +1,200 @@
...
+
+	pushDetails, err := parseWebhookPushDetails(c, body)
+	if err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "invalid webhook payload"})
+		return
+	}
+
+	now := time.Now()
+	if _, err := gorm.G[models.Repository](db.DB).Where("id = ?", id).Updates(c.Request.Context(), models.Repository{
+		SyncStatus:    models.SyncStatusSuccess,
```
alex289, line 86: Im not sure. Why is the status unkown when creating an repository? Isn't this confusing?

alex289 (approved):
I can verify gitea also works

backend/internal/hub/routes/webhooks.go
timokoessler: I think it is the safest default? Also not changed by this PR. Otherwise you might thing everything is working although you never set up the webhook maybe?

PR #117, opened 2026-04-22
test: initial e2e setup

timokoessler:


e2e/package.json
```diff
@@ -0,0 +1,18 @@
...
+		"test:ui": "playwright test --headed --ui",
+		"install:browsers": "playwright install"
+	},
+	"devDependencies": {
+		"@playwright/test": "^1.59.1",
+		"@types/node": "^25.6.0",
+		"typescript": "^6.0.3"
+	},
+	"engines": {
+		"node": ">=24.0.0"
```
alex289, line 16: We use 25 in docker and ci I think. Maybe this should all be the same

alex289: Ah we use 24 in ci. This seems very inconsistent

timokoessler: You updated some stuff to v25 :D, but will change.

timokoessler: This is only the minimum version tho, safe too keep v24, no differences that affect us.

alex289: > You updated some stuff to v25 :D, but will change.

I knooo :D. Dependabot kept annoying me

alex289 (approved):

PR #121, opened 2026-04-24
feat: Add reset password command

alex289:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

Add reset-password command to hub:

- Receives user id or email
- Sets a random password equal to the admin user update endpoint
- Show info to user


backend/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.26.1
 require (
 	github.com/aegis-aead/go-libaegis v0.2.14
 	github.com/alexedwards/argon2id v1.0.1-0.20251028180742-493d7dead70e
+	github.com/charmbracelet/lipgloss v1.1.0
```
timokoessler: Why not v2?

backend/cmd/hub/reset_password.go
```diff
@@ -0,0 +1,193 @@
...
+		logLevel = parsedLogLevel
+	}
+
+	isDemoMode, _ := strconv.ParseBool(os.Getenv("DEMO"))
+	if isDemoMode {
+		return errors.New("password reset is not available in demo mode")
+	}
+
+	dbLogger := zerolog.New(io.Discard)
+	if err := db.Connect(dbLogger, logLevel, false); err != nil {
```
timokoessler, line 67: I read https://www.sqlite.org/faq.html#q5 and this should normally work. I am still thinking whether it would make sense to require the user to stop the app for password reset, also as a "security feature", but also to make db corruption less likely. I read that SQLite's default is to fail immediately if locked, so I guess if the hub is executing some queries at the same time, they will crash. Also I read that the locking might not work with Docker Desktop because of the VM. Maybe we can open the db with the following connection string parameters and a better error messages that tells the user to stop the application first? `?_locking_mode=exclusive&_busy_timeout=1`

alex289: I tried the connection string but it didn't work because when running the hub every query was showing some errors somehow. I set some some other flags and a pool to try to reduce the risk of corruption. Hope this is okay

backend/internal/hub/db/db.go
```diff
@@ -19,6 +19,31 @@ var migrationFiles embed.FS
...
+
+func configureSQLitePool(db *gorm.DB) error {
+	sqlDB, err := db.DB()
+	if err != nil {
+		return err
+	}
+
+	// SQLite allows a single writer; limiting the pool to one connection avoids
+	// internal lock contention from multiple pooled connections.
+	sqlDB.SetMaxOpenConns(1)
```
timokoessler, line 40: We could have multiple readers in WAL mode tho (the main advantage of WAL mode)?

backend/internal/hub/db/db.go
```diff
@@ -19,6 +19,31 @@ var migrationFiles embed.FS
 
 var DB *gorm.DB
 
+const sqliteFilePath = "data/hub.db"
+
+func sqliteReadWriteDSN() string {
+	return sqliteFilePath + "?_busy_timeout=5000&_journal_mode=WAL&_synchronous=NORMAL&_foreign_keys=ON"
```
timokoessler, line 25: Great change, we need to update backup instructions in docs tho to backup the whole volume.

timokoessler (approved):

backend/internal/hub/db/db.go
alex289: So should I remove wal?

timokoessler: No, but maybe we could increase open connections to e.g. 10? Discussed it a bit with AI, but this might indeed be a bit safer.

timokoessler: In case you decide to increase the idle also needs to be increased.

PR #135, opened 2026-04-28
feat: Add backup cmd to CLI

timokoessler:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

Adds a backup command for full db backups while the app is running.


alex289:
Don't we need a backup command for the agent too so we can backup the compose files etc.?

alex289 (approved):
Niceeee

timokoessler:
> Don't we need a backup command for the agent too so we can backup the compose files etc.?

Maybe at some point, although we have the compose files in hub db, so it is mainly about data volumes which can become very large ofc.

PR #146, opened 2026-05-07
feat: Add agent sync

maximbigler:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

Fixes #33

This PR introduces the initial implementation of the agent sync. It copies docker-compose.yml into the agent volume, and sets the application status accordingly.


backend/internal/hub/websocket/hub.go
```diff
@@ -30,6 +30,8 @@ type Hub struct {
 	log     *zerolog.Logger
 }
 
+var DefaultHub *Hub
```
alex289: Why is this needed?

backend/internal/agent/deploy.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	composePath := filepath.Join(appDir, "docker-compose.yml")
+	if err := os.WriteFile(composePath, []byte(content), 0640); err != nil {
+		return "", err
+	}
+	return composePath, nil
+}
+
+func runDockerCompose(ctx context.Context, args ...string) (string, error) {
+	cmd := exec.CommandContext(ctx, "docker", append([]string{"compose"}, args...)...)
```
alex289: This should use the docker compose sdk right? @timokoessler

backend/internal/agent/websocket.go
```diff
@@ -95,7 +96,23 @@ func sendMessage(conn *websocket.Conn, session *wscrypto.Session, msg *messages.
 	return conn.WriteMessage(websocket.BinaryMessage, data)
 }
 
+type wsSender struct {
+	mu      sync.Mutex
+	conn    *websocket.Conn
+	session *wscrypto.Session
+}
+
+func (s *wsSender) Send(msg *messages.ClientMessage) error {
```
alex289: We already had a send message method above right?

backend/internal/agent/deploy.go
timokoessler: Yes, that is why we are using it :D. There is also no Docker Command or even a shell available in the docker image! It is already installed and initialized. See https://docs.docker.com/compose/compose-sdk/

backend/internal/agent/deploy.go
```diff
@@ -0,0 +1,167 @@
...
+		ApplicationId: req.ApplicationId,
+		SyncStatus:    "synced",
+		HealthStatus:  healthStatus,
+		Message:       "deployment applied",
+		ReportedAt:    now,
+	}
+}
+
+func writeComposeFile(appID, content string) (string, error) {
+	appDir := filepath.Join("data", "applications", appID)
```
timokoessler: I would suggest adding another volume to the docker compose file, so user can choose to move the apps somewhere else than inside the data dir? To discuss: maybe use human readable names instead of id? Then we have to enforce unique names tho.

backend/internal/hub/applications/sync.go
```diff
@@ -78,9 +80,29 @@ func processSyncJob(ctx context.Context, job syncJob, log *zerolog.Logger) {
...
+		},
+	}
+
+	if sent := hubws.DefaultHub.Send(job.Application.AgentId, deployMsg); !sent {
+		log.Error().Str("applicationId", job.Application.Id).Str("agentId", job.Application.AgentId).
+			Msg("failed to trigger deployment: agent is offline or busy")
+		return
+	}
 
 	if _, err := gorm.G[models.Application](db.DB).
```
timokoessler: This branch would also need success = true?

backend/internal/agent/deploy.go
alex289: We could do something like "provider-owner-repo" e.g. "github-orcacd-docs" maybe

.env.example
```diff
@@ -8,3 +8,4 @@ APP_SECRET= # Use: openssl rand -base64 42
 HUB_URL=https://... # Public URL of the Hub, should be the same as APP_URL
 
 AUTH_TOKEN= # Create a token in the Hub UI and use it here
+DEPLOYMENTS_DIR=/app/deployments # Directory where the agent stores compose files before running them
```
alex289: Shouldn't we be just deployments and we prefix the app ourselfs? @timokoessler

alex289: What use case is there to make this customizable?

backend/internal/agent/docker/deploy.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	if req.ComposeFile == "" {
+		return errors.New("compose file is empty")
+	}
+
+	applicationDir := filepath.Join(c.deploymentsDir, req.ApplicationName)
+	composePath := filepath.Join(applicationDir, composeFileName)
+
+	if err := os.MkdirAll(applicationDir, 0o750); err != nil {
```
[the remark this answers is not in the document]
timokoessler: Add a shared helper like "doesNotLookLikeFilePath"?

backend/internal/agent/docker/deploy.go
```diff
@@ -0,0 +1,90 @@
...
+		return fmt.Errorf("create deployment directory: %w", err)
+	}
+	if err := os.WriteFile(composePath, []byte(req.ComposeFile), 0o600); err != nil {
+		return fmt.Errorf("write compose file: %w", err)
+	}
+
+	project, err := loadProject(ctx, c.compose, api.ProjectLoadOptions{
+		ProjectName: "orca-" + strings.ToLower(req.ApplicationName),
```
timokoessler: Not sure if we want to prepend orca? Maybe add a label to all containers managed by OrcaCD?

backend/internal/agent/agent.go
```diff
@@ -31,13 +31,14 @@ type FatalConfigError struct {
...
-	HubPublicKey ed25519.PublicKey
-	HealthPort   string
+	LogLevel       zerolog.Level
+	LogJSON        bool
+	HubUrl         string
+	AuthToken      string
+	AgentID        string
+	HubPublicKey   ed25519.PublicKey
+	HealthPort     string
+	DeploymentsDir string
```
timokoessler, line 41: We discussed in real-life that it is not needed, but I guess we need it for development?

backend/internal/agent/websocket.go
```diff
@@ -7,15 +7,46 @@ import (
...
+
+func newMessageSender(conn *websocket.Conn, session *wscrypto.Session) *messageSender {
+	return &messageSender{
+		conn:    conn,
+		session: session,
+	}
+}
+
+func (s *messageSender) Send(msg *messages.ClientMessage) error {
+	s.mu.Lock()
```
timokoessler: We don't need a mutex for the whole encryption part I guess? Maybe a todo for later, that we should make the send logic async by adding it to a goroutine?

backend/internal/hub/applications/sync.go
```diff
@@ -78,23 +78,36 @@ func processSyncJob(ctx context.Context, job syncJob, log *zerolog.Logger) {
...
+		log.Error().Str("applicationId", job.Application.Id).Msg("agent deployment finished without a result")
+		return
+	}
+	if !result.Success {
+		log.Error().
+			Str("applicationId", job.Application.Id).
+			Str("request_id", result.RequestId).
+			Str("error", result.ErrorMessage).
+			Msg("agent reported deployment failure")
+		markDeploymentExecutionFailure(ctx, job.Application.Id, log)
```
timokoessler: defer func() will still be called because success = false

backend/internal/hub/routes/applications.go
```diff
@@ -271,6 +272,54 @@ func DeleteApplicationHandler(c *gin.Context) {
...
+	if err != nil {
+		if errors.Is(err, hubApplications.ErrAgentUnavailable) {
+			c.JSON(http.StatusConflict, gin.H{"error": "agent is not connected"})
+			return
+		}
+		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to start deployment"})
+		return
+	}
+
+	if _, err := gorm.G[models.Application](db.DB).
```
timokoessler: Not sure if it makes more sense to let the deployer handle status updates?

backend/internal/hub/websocket/handler.go
```diff
@@ -167,6 +167,13 @@ func handleClientMessage(client *Client, msg *messages.ClientMessage, log *zerol
 		if err != nil {
 			log.Error().Err(err).Str("client", client.Id).Msg("Failed to update last_seen")
 		}
+	case *messages.ClientMessage_DeployResult:
+		if !h.ResolveDeploy(p.DeployResult) {
```
timokoessler, line 180: Is this updating the db status?

backend/internal/hub/websocket/hub.go
```diff
@@ -19,19 +21,61 @@ type Client struct {
...
 }
 
 type Hub struct {
-	mu      sync.RWMutex
-	clients map[string]*Client
-	log     *zerolog.Logger
+	mu             sync.RWMutex
+	clients        map[string]*Client
+	deploysMu      sync.Mutex
+	pendingDeploys map[string]pendingDeploy
```
timokoessler: We are mixing websocket and deployment state, maybe add to a separate file?

.env.example
timokoessler: E.g. users wanting to mount all compose files in /opt.

docker-compose.dev.yml
```diff
@@ -20,11 +20,14 @@ services:
...
     container_name: orca-agent-dev
     use_api_socket: true # Allows container to access Docker API on host
+    volumes:
+      - ./data/agent:/app/deployments
     depends_on:
       - hub
     environment:
       - HUB_URL=ws://hub:8080/api/v1/ws
       - AUTH_TOKEN=
+      - DEPLOYMENTS_DIR=/app/deployments
```
timokoessler: Maybe only /deployments

docker-compose.yml
```diff
@@ -23,5 +23,7 @@ services:
     cap_drop:
       - ALL
     use_api_socket: true # Allows container to access Docker API on host
+    volumes:
+      - ./data/agent:/app/deployments
```
timokoessler: Don't mount in data?

backend/internal/hub/websocket/handler.go
maximbigler: No, it's not updating the database. The ResolveDeploy call only completes the in-memory deployment handle

docker-compose.yml
```diff
@@ -23,5 +23,7 @@ services:
     cap_drop:
       - ALL
     use_api_socket: true # Allows container to access Docker API on host
+    volumes:
+      - ./data/agent/deployments:/deployments
```
timokoessler:
```suggestion
      - ./deployments:/deployments
```
Maybe? You can the backup data volume to backup OrcaCD without all the application data of different apps.

backend/internal/shared/utils/fileUtils.go
```diff
@@ -0,0 +1,53 @@
...
+type FileUtils struct{}
+
+// DoesNotLookLikeFilePath validates that a path is safe and doesn't contain path traversal sequences.
+func DoesNotLookLikeFilePath(name string) error {
+	if strings.TrimSpace(name) == "" {
+		return errors.New("file path cannot be empty")
+	}
+
+	// Prevent path traversal attacks
+	if strings.Contains(name, "..") {
```
timokoessler:
```suggestion
	if strings.Contains(name, "/..") || strings.Contains(name, "../") {
```
To reduce false positives? .. is not dangerous as part of app name, although very unlikely.

backend/internal/hub/websocket/handler.go
timokoessler: I still don't get where the db status is updated when client sends result :D

backend/internal/agent/websocket.go
```diff
@@ -7,15 +7,47 @@ import (
...
+		session: session,
+	}
+}
+
+// TODO: We should make this async by adding it to a goroutine
+func (s *messageSender) Send(msg *messages.ClientMessage) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	return sendMessage(s.conn, s.session, msg)
```
timokoessler: You are correct that conn.WriteMessage needs a mutex. So we should prevent usage of and refactor existing usage of sendMessage?

backend/internal/agent/agent.go
timokoessler, line 41: Add to development example env with relative dir so it works outside of container in dev?

backend/internal/agent/docker/deploy.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return fmt.Errorf("load compose project: %w", err)
+	}
+
+	// Add OrcaCD managed label to all services
+	for _, service := range project.Services {
+		if service.Labels == nil {
+			service.Labels = make(map[string]string)
+		}
+		service.Labels["managed_by"] = "orca-cd"
```
timokoessler, line 85: Cool! :)

backend/internal/agent/agent.go
alex289: Isn't this optional?

timokoessler: The env var, yes, but in dev it will try to write in /deployments on root of hard drive.

backend/internal/shared/utils/fileUtils.go
maximbigler: Yeah nice catch! :D

backend/internal/agent/agent.go
maximbigler: You now can config the deployment dir, the default is /deployments and I added a example

backend/internal/agent/websocket.go
maximbigler: Yep, we need mutex, because the conn.WriteMessage is not thread-safe. Here the documentation for that: https://pkg.go.dev/github.com/gorilla/websocket#hdr-Concurrency :D

backend/internal/hub/websocket/handler.go
maximbigler: Yeah I looked at the code and yes it is complicated. In here I wait for the pending Deploy, which will be trigger by the code you mentioned: 
https://github.com/OrcaCD/orca-cd/pull/146/changes#diff-5e2a16ef0a3456802c8862ba523e73f8a9614001ca6d7ca81c9e435efcbb24c0R33
The application deploy.go in the hub wait's for the agent to give a status:
https://github.com/OrcaCD/orca-cd/pull/146/changes#diff-47d9c9b7a44a81299bb8fe665de47fbcd8e210e162a28c09737fd2fc3a2ad2abR117
I found one probleme: The pendingDeploys are in-memory, which means if the hub crashes the status updates will fail on running deploys. I already added a TODO for that, I would suggest to fix it on another pr. :D
We may want to simplify the process.

maximbigler: This will await and stopps the thread unitill it recived a message from the channel:
`:= <-h.outcome`

alex289 (approved):

alex289 (approved):

PR #161, opened 2026-05-12
feat: Notifications

### LINKED ISSUE #112: feat: Configurable notifications
We want to use Shouterrr. To discuss: Maybe add an events table?

alex289:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

Fixes #112 

- Add shouterrr
- Add notifications to db
- Add crud endpoints
- Add tests
- Add frontend


backend/internal/hub/routes/notifications.go
```diff
@@ -0,0 +1,433 @@
...
+	"gorm.io/gorm"
+)
+
+const NotificationsPath = "/api/v1/notifications"
+
+type createNotificationRequest struct {
+	Name            string                    `json:"name" binding:"required,min=1,max=128"`
+	Enabled         *bool                     `json:"enabled"`
+	EnableByDefault *bool                     `json:"enableByDefault"`
+	Status          models.NotificationStatus `json:"status" binding:"required"`
```
timokoessler: Why is status on creation sent by client?

backend/internal/hub/routes/notifications.go
```diff
@@ -0,0 +1,433 @@
...
+	Type            models.NotificationType   `json:"type" binding:"required"`
+	Config          json.RawMessage           `json:"config" binding:"required"`
+	ApplicationIds  []string                  `json:"applicationIds"`
+}
+
+type updateNotificationRequest struct {
+	Name            string                    `json:"name" binding:"required,min=1,max=128"`
+	Enabled         *bool                     `json:"enabled"`
+	EnableByDefault *bool                     `json:"enableByDefault"`
+	Status          models.NotificationStatus `json:"status" binding:"required"`
```
timokoessler: Same as above.

backend/internal/hub/routes/notifications.go
```diff
@@ -0,0 +1,433 @@
...
+
+const NotificationsPath = "/api/v1/notifications"
+
+type createNotificationRequest struct {
+	Name            string                    `json:"name" binding:"required,min=1,max=128"`
+	Enabled         *bool                     `json:"enabled"`
+	EnableByDefault *bool                     `json:"enableByDefault"`
+	Status          models.NotificationStatus `json:"status" binding:"required"`
+	Type            models.NotificationType   `json:"type" binding:"required"`
+	Config          json.RawMessage           `json:"config" binding:"required"`
```
timokoessler: It might be possible to simplify the logic a lot by always using json for the config and always decode it just once and not in route handler and providers?

alex289: Im not sure what you mean by this

timokoessler: I think there are multiple places in the code where it checks if it's json and also allows non json? E.g. in parseDirectTargets in common.go?

alex289: I tried improving it. Only the providers are parsing now which is necessary. Good enough?

frontend/package.json
```diff
@@ -13,6 +13,7 @@
 		"translations:check": "inlang validate --project ./project.inlang"
 	},
 	"dependencies": {
+		"@base-ui/react": "^1.4.1",
```
timokoessler, line 16: The project is using radix ui?

alex289: Unfortunately the shadcn combobox does only provide a base-ui implementation

timokoessler: Interesting 😅

timokoessler: Maybe switch at some point :D

frontend/src/components/dialogs/upsert-notification.tsx
timokoessler, line 1: Image
The modal is a bit large, long-term we can maybe do two steps or collapse additional optional settings, although it's also not nice to hide them.

timokoessler: The avatar and bot name can also be configured in Discord UI (with image upload). Not sure which of both settings has a higher priority? Maybe enough to just let the user configure them in Discord?

frontend/src/components/dialogs/upsert-notification.tsx
```diff
@@ -0,0 +1,377 @@
...
+			throw new Error(m.validationNotificationWebhookUrlInvalid());
+		}
+
+		return config;
+	}
+
+	return "";
+}
+
+export default function UpsertNotificationDialog({
```
timokoessler: In edit mode the dialog closes itself directly when the cursor moves away from the dropdown (at least in Firefox)

backend/internal/hub/notifications/provider/discord.go
```diff
@@ -0,0 +1,69 @@
...
+	URLs      []string `json:"urls"`
+}
+
+func (DiscordProvider) BuildShouterrrUrls(rawConfig string) ([]string, error) {
+	trimmed, err := normalizeRawConfig(rawConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	if strings.HasPrefix(trimmed, "{") {
```
timokoessler: We always send JSON from frontend to backend, or?

frontend/src/components/dialogs/upsert-notification.tsx
alex289: Splitting is a good idea. I would keep the name and avatar. Users can use the same webhook for multiple services which then all set the name and avatar in their own messages

frontend/src/components/dialogs/upsert-notification.tsx
alex289: I removed editing for now. It's not clean like this with returning the config to the client

timokoessler (approved):

PR #173, opened 2026-06-01
feat: Rename backup command to export

maximbigler:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

Renamed the backup command to export command


timokoessler:
For me an export is more something I can use in any way, and an import something that does not override existing data, like a backup restore?

backend/cmd/hub/export.go
```diff
@@ -16,34 +16,34 @@ import (
...
+func newExportCmd() *cobra.Command {
 	var outputPath string
 
 	cmd := &cobra.Command{
-		Use:   "backup",
-		Short: "Create a backup of the hub database",
-		Long:  "Create a consistent backup of the hub database using VACUUM INTO. Safe to run while the hub is running.",
+		Use:   "export",
+		Short: "Create a export of the hub database",
+		Long:  "Create a consistent export of the hub database using VACUUM INTO. Safe to run while the hub is running.",
```
alex289: This should still be backup for clarification maybe?

alex289 (approved):

alex289:
> For me an export is more something I can use in any way, and an import something that does not override existing data, like a backup restore?

PocketID has a similar approach

https://pocket-id.org/docs/setup/data-export-import

alex289 (approved):

Issue #175, opened 2026-06-02
feat: Notification extensions


alex289:
Add notification providers:
- [x] Generic Webhooks
- [x] Slack
- [ ] Email
- [x] MS Teams
- [x] Gotify
- [x] Shouterrr url?

Send notifications
- [x] Successful deployment
- [x] Failed deployment
- [x] Repo sync failed

Also add icons to the notifications list

PR #183, opened 2026-06-04
feat: Webhook notification provider

alex289:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

Add generic webhook notification provider.

Part of #175


backend/internal/hub/notifications/provider/webhook.go
```diff
@@ -0,0 +1,137 @@
...
+
+type WebhookProvider struct{}
+
+type webhookConfig struct {
+	WebhookURL string            `json:"webhookUrl"`
+	Method     string            `json:"method"`
+	Headers    map[string]string `json:"headers"`
+}
+
+var supportedWebhookMethods = []string{
```
timokoessler: Do we really want to allow the user to send a GET request with HTTP body or DELETE? Maybe force users to normal API design? 😄

alex289: I thought about that too and was unsure. Zerobyte lets users do that too and even set a custom content type even though json is sent. But I can remove it sure

timokoessler: We can always add again if a user requests it and has a really good reason?

timokoessler (approved):

PR #205, opened 2026-06-19
feat: Key-Rotate command

### LINKED ISSUE #167: feat: Key rotate command
We want to add command which allows users to rotate their encryption key. PocketID has a similar command which we could get inspiration from

alex289:
## Summary
- Add a new `hub key rotate` command to re-encrypt database fields from an old `APP_SECRET` to the current one.
- Cover agents, applications, repositories, notifications, and OIDC providers, with confirmation prompts and safe validation for secret length and equality.
- Surface a concise success summary after rotation and block execution in demo mode or when the database is unavailable.
- Add tests for rotation behavior, secret validation, and command wiring.

## Testing
- Added unit tests covering successful re-encryption, invalid secret rejection, and CLI command registration.
- Not run (not requested).

Fixes #167


backend/cmd/hub/key_rotate.go
```diff
@@ -0,0 +1,341 @@
...
+
+	if err := validateKeyRotateSecrets(options.OldSecret, cfg.AppSecret); err != nil {
+		return err
+	}
+
+	warningPoints := []string{
+		"The hub server must be stopped before rotating the key",
+		"Create a database export before rotating the key",
+		"The database will be re-encrypted with the current APP_SECRET",
+		"Sessions and agent tokens signed with the old APP_SECRET may need to be re-issued",
```
timokoessler: Remove `may`?

backend/cmd/hub/key_rotate.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+	if oldSecret == newSecret {
+		return errSameSecrets
+	}
+
+	return nil
+}
+
+func rotateDatabaseEncryptionKey(ctx context.Context, oldSecret, newSecret string) (keyRotationResult, error) {
+	if err := validateKeyRotateSecrets(oldSecret, newSecret); err != nil {
```
timokoessler, line 176: Is this intentionally called twice?

backend/cmd/hub/key_rotate.go
```diff
@@ -0,0 +1,341 @@
...
+		}
+		return fmt.Errorf("failed to connect to database: %w", err)
+	}
+	defer func() {
+		_ = db.Close()
+	}()
+
+	result, err := rotateDatabaseEncryptionKey(ctx, options.OldSecret, cfg.AppSecret)
+	if err != nil {
+		return fmt.Errorf("key rotation failed: %w", err)
```
timokoessler: Add what to do next? E.g. restore old database? Maybe set some flag in db that a key rotation process is pending, so in case it failed before we know that? Might be a bit complex tho.

backend/cmd/hub/key_rotate.go
```diff
@@ -0,0 +1,341 @@
...
+
+func rotateEncryptedBatches[T any](
+	ctx context.Context,
+	tx *gorm.DB,
+	oldSecret string,
+	newSecret string,
+	label string,
+	rotate func([]T) error,
+) error {
+	if err := crypto.Init(oldSecret); err != nil {
```
timokoessler: Maybe we can refactor crypto so it is possible to create instances and have a Default one, like for websockets or in other parts of the code? This would still not be nice, but maybe we just have to switch default instance then.

backend/cmd/hub/key_rotate.go
```diff
@@ -0,0 +1,341 @@
...
+		}
+		return nil
+	})
+}
+
+func rotateAgents(ctx context.Context, tx *gorm.DB, agents []models.Agent, result *keyRotationResult) error {
+	for _, agent := range agents {
+		rows, err := gorm.G[models.Agent](tx).
+			Where("id = ?", agent.Id).
+			Updates(ctx, models.Agent{Name: agent.Name, KeyId: agent.KeyId})
```
timokoessler: As soon as we add encrypted fields and forget to add them, the rotate will be broken. AI told me the following (not verified or tested):


You can use Go reflection combined with GORM's own schema parser to discover EncryptedString fields dynamically. Here's the approach:

```go
  var (
      encryptedType    = reflect.TypeOf(crypto.EncryptedString(""))
      encryptedPtrType = reflect.TypeOf((*crypto.EncryptedString)(nil))
  )

  // Returns map of table → []column for all EncryptedString fields
  func findEncryptedColumns(db *gorm.DB, models []any) (map[string][]string, error) {
      result := make(map[string][]string)
      for _, m := range models {
          stmt := &gorm.Statement{DB: db}
          if err := stmt.Parse(m); err != nil {
              return nil, err
          }
          for _, field := range stmt.Schema.Fields {
              ft := field.FieldType
              if ft == encryptedType || ft == encryptedPtrType {
                  result[stmt.Schema.Table] = append(result[stmt.Schema.Table], field.DBName)
              }
          }
      }
      return result, nil
  }
```
  You'd call it with all your model structs:

```go
  allModels := []any{
      &models.Agent{},
      &models.Application{},
      &models.Repository{},
      &models.OIDCProvider{},
  }
  cols, err := findEncryptedColumns(db, allModels)
```

The key thing stmt.Parse buys you is that field.DBName respects gorm:"column:..." tags (like OIDCProvider.ClientSecret which has column:client_secret explicitly set), so you get the actual DB column names rather than the Go field names lowercased.
The only thing you can't avoid is maintaining the allModels slice — GORM has no global model registry. But since you control all model files, it's a one-liner addition whenever a new model is added.

For the key rotation itself, the flow would be: for each {table, columns} pair, SELECT id, <cols...> FROM <table> (GORM will auto-decrypt via Scan), then UPDATE each row (GORM will re-encrypt with the new key via Value). The EncryptedString valuer/scanner handles the crypto transparently.

timokoessler (commented):
Cool

timokoessler (approved):

PR #208, opened 2026-06-22
feat: Refactor agent and hub sync for better readability

### LINKED ISSUE #172: fix: Sync issues
- [x] Sync app when application gets created
- [x] No error message in ui for failed deployments
- [x] Lock sync button when already syncing
- [x] Check if sync buttons work and check what they sync
- [x] Problems when agent or hub stop during syncs
- [x] Check TODO comments
- [x] Duplicate app name
- [x] Delete the application also on agent side

### LINKED ISSUE #211: Normalize the compose name and skip check
What if we always just normalize the name and skip the check? I think it replaces all special chars. So then we could remove DoesNotLookLikeFilePath at some point. (But maybe in another PR :D)

_Originally posted by @timokoessler in https://github.com/OrcaCD/orca-cd/pull/208#discussion_r3485857832_

maximbigler:
## **Type of change**

- [ ] 🐛 Bug fix
- [ ] 🚀 New feature
- [x] ❓ Other (please specify): Refactor

## **Description**

Fixes #172
Fixes #211

Refactors the repository-sync and application-deploy flow in `internal/hub/applications` to make it readable and to remove duplicated/divergent logic. No new feature; behavior is preserved except for the intentional changes called out below.

### Application-level deploy state machine

- Introduced `updateApplicationStatus` as the single chokepoint for every application status write + SSE publish; the four `mark*` helpers are now thin wrappers over it. Fixes a bug where `markDeploymentInProgress` published an SSE event twice on a DB error.
- Extracted `applyDeployResult` (+ `deployNotifications`) — one shared interpreter of an agent's deploy outcome (`err` / nil result / `!Success` / success), replacing two divergent copies that previously lived in the sync path and the manual-deploy path.
- Split `Deployer.StartDeploy` into `dispatchDeploy` (transport-only) and `StartDeploy` (dispatch + mark in-progress). The sync path uses `dispatchDeploy` via `DeployAndWait` because `processSyncJob` already owns the in-progress transition, so each path now marks the application in-progress exactly once instead of twice.
- Rewrote `processSyncJob` to use linear control flow instead of the `success bool` + `defer` pattern.
- **Bug fix:** terminal status writes now use `context.Background()` instead of the job context. A deploy that hit the 3-minute job timeout previously wrote its failure status on an already-cancelled context, which could leave the application stuck displaying `Syncing`.

### Unified sync entry points

- Added `SyncApplications`, the single path that all four sync triggers funnel through (polling, manual repo sync, webhooks, GitHub Actions). It performs consistent repository-status bookkeeping (`Syncing → Success/Failed`).
- Added `CommitResolver` with `StaticCommit` (webhooks and GitHub Actions already carry the pushed SHA) and `LatestCommit` (polling and commit-less generic webhooks), separating "where the commit comes from" from the shared sync logic.
- `SyncRepository`, `WebhookHandler`, `handleGenericWebhook`, and the GitHub Actions handler now delegate to `SyncApplications`; the bespoke `enqueueGenericApps` helper and the inline repository-status updates were removed.

### Documentation

- Doc comments make the two-layer model explicit (repository-level sync vs. application-level deploy) and clarify that a repository's `SyncStatus` reflects whether the sync was *dispatched* (commits resolved, jobs enqueued), not whether every application finished deploying — per-application progress lives on each `Application`.

### Intentional behavior changes (please review)

1. A sync-path transport failure (agent dropped / nil result) now marks the application `OutOfSync` only, instead of also `Unhealthy` — matching the manual-deploy path. A dropped connection does not mean the running containers are unhealthy.
2. Webhooks and GitHub Actions now perform the full `Syncing → Success/Failed` repository bookkeeping like the poller. Previously webhooks marked `Success` only (never `Failed`) and GitHub Actions did no repository-status bookkeeping at all. For these synchronous handlers the `Syncing` state is effectively instantaneous; it is only visibly observable on the polling path during the network commit lookup.
3. Generic webhooks now skip applications with an empty branch (consistent with the poller) and mark the repository `Failed` if commit resolution fails, instead of silently enqueuing an empty commit.

### Out of scope (planned follow-up)

Removing the `Default*` package-level globals in favor of dependency injection, which also requires reordering the `server.go` bootstrap. Deliberately left for a separate PR to keep this diff focused and reviewable.

### Testing

`applications`, `routes`, and `websocket` suites pass; `gofmt` / `go vet` clean. The only failing tests are the Docker-daemon-dependent ones in `internal/agent/docker`, which this branch does not touch.


alex289:
Does somebody has credits for copilot review? I think this is appropriate for this pr 😂

frontend/src/routes/_authenticated/applications/$id.index.tsx
```diff
@@ -233,6 +235,16 @@ function ApplicationDetailsPage() {
 				</BreadcrumbList>
 			</Breadcrumb>
 
+			{data?.lastSyncError && (
+				<Alert variant="destructive" className="border-destructive/50 bg-destructive/10">
```
timokoessler: We already have a component for that: https://github.com/OrcaCD/orca-cd/blob/main/frontend/src/components/alerts/error-alert.tsx

backend/internal/hub/websocket/handler.go
```diff
@@ -108,6 +108,10 @@ func WsHandler(h *Hub, log *zerolog.Logger) gin.HandlerFunc {
 		if err != nil {
 			log.Error().Err(err).Str("agent_id", claims.Subject).Msg("Failed to update status to online")
 		}
+		_, applicationErr := gorm.G[models.Application](db.DB).Where("agent_id = ?", claims.Subject).Update(c.Request.Context(), "health_status", models.Healthy)
```
timokoessler: Not all apps are automatically healthy then the agent connects?

backend/internal/hub/websocket/delete.go
```diff
@@ -0,0 +1,72 @@
...
+// never loses track of containers when the agent is offline or the removal fails.
+func (h *Hub) RemoveApplication(ctx context.Context, agentID, applicationID, applicationName string) (*messages.DeleteResult, error) {
+	requestID := uuid.NewString()
+	ch := make(chan *messages.DeleteResult, 1)
+
+	h.deleteMu.Lock()
+	h.pendingDeletes[requestID] = ch
+	h.deleteMu.Unlock()
+
+	cleanup := func() {
```
timokoessler: Use defer? Expected that it's not always called right now?

backend/internal/hub/models/applications.go
```diff
@@ -37,6 +37,7 @@ type Application struct {
 	Commit                   string                  `gorm:"type:text;not null"`
 	CommitMessage            string                  `gorm:"type:text;not null"`
 	LastSyncedAt             *time.Time              `gorm:"type:timestamp;"`
+	LastSyncError            *string                 `gorm:"type:text;"`
```
alex289, line 49: Im on the phone so I can't search here. Where is this set?

backend/internal/hub/websocket/deploy.go
```diff
@@ -2,132 +2,58 @@ package websocket
...
-	dm.mu.Lock()
-	pending, ok := dm.pendingDeploys[result.RequestId]
-	if ok {
-		delete(dm.pendingDeploys, result.RequestId)
-	}
-	dm.mu.Unlock()
-
-	if !ok {
-		return false
+		notifications.SendNotification(result.ApplicationId, "Success: deployment succeeded for "+result.ApplicationId, log)
```
timokoessler: Use name instead of id in notification.

backend/internal/hub/websocket/deploy.go
```diff
@@ -2,132 +2,58 @@ package websocket
...
+	errMsg := result.ErrorMessage
+	if errMsg == "" {
+		errMsg = "deployment failed"
 	}
-	dm.mu.Unlock()
 
-	if ok {
-		close(pending.outcome)
-	}
+	notifications.SendNotification(result.ApplicationId, "Error: deployment failed for "+result.ApplicationId, log)
```
timokoessler: Keep name in notification instead of id.

backend/internal/hub/models/applications.go
timokoessler: https://github.com/OrcaCD/orca-cd/pull/208/changes#diff-cc13394d4d0b5ae0dfcc39dde05b58ec8cb3f176b34860bc892ad13180a4dcf3R66

backend/internal/agent/docker/deploy.go
```diff
@@ -142,3 +151,46 @@ func (c *Client) Deploy(ctx context.Context, req DeployRequest) error {
...
+		return errors.New("docker compose service is not initialized")
+	}
+	if c.deploymentsDir == "" {
+		return errors.New("deployments directory is not configured")
+	}
+	if !c.Ready() {
+		return errors.New("docker daemon is not ready")
+	}
+
+	if err := utils.DoesNotLookLikeFilePath(req.ApplicationName); err != nil {
```
timokoessler, line 168: What if we always just normalize the name and skip the check? I think it replaces all special chars. So then we could remove DoesNotLookLikeFilePath at some point. (But maybe in another PR :D)

backend/internal/hub/applications/queue.go
```diff
@@ -33,6 +33,14 @@ func NewQueue(log *zerolog.Logger) *Queue {
 	}
 }
 
+func QueueLogger() *zerolog.Logger {
+	if DefaultQueue != nil {
+		return DefaultQueue.log
+	}
+	nop := zerolog.Nop()
```
timokoessler: Isn't the logger always available?

backend/internal/hub/routes/applications.go
```diff
@@ -131,6 +135,27 @@ func GetApplicationHandler(c *gin.Context) {
 	c.JSON(http.StatusOK, toApplicationResponse(&application))
 }
 
+// applicationNameTaken reports whether another application already uses name
+// (case-insensitive). Names are stored encrypted with a random nonce, so a DB
+// unique index can't enforce this — we decrypt and compare in code. excludeID
+// skips a record (the one being updated).
+func applicationNameTaken(ctx context.Context, name, excludeID string) (bool, error) {
```
timokoessler: Alternative approach would be to add a new column that contains a hash of the name maybe?

backend/internal/hub/routes/applications.go
```diff
@@ -131,6 +135,27 @@ func GetApplicationHandler(c *gin.Context) {
...
+	apps, err := gorm.G[models.Application](db.DB).Find(ctx)
+	if err != nil {
+		return false, err
+	}
+	target := strings.TrimSpace(name)
+	for i := range apps {
+		if apps[i].Id == excludeID {
+			continue
+		}
+		if strings.EqualFold(strings.TrimSpace(apps[i].Name.String()), target) {
```
timokoessler: Shouldn't it be possible to have a app with the same name but on another agent?

backend/internal/hub/routes/applications.go
```diff
@@ -369,19 +443,15 @@ func DeployApplicationHandler(c *gin.Context) {
...
+
 	if err != nil {
-		if errors.Is(err, hubApplications.ErrAgentUnavailable) {
-			c.JSON(http.StatusConflict, gin.H{"error": "agent is not connected"})
-			return
-		}
-		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to start deployment"})
+		c.JSON(http.StatusInternalServerError, gin.H{"error": fmt.Sprintf("failed to trigger application deploy: %v", err)})
 		return
+	} else {
```
timokoessler: This else is not needed or?

alex289 (approved):

backend/internal/hub/applications/sync.go
```diff
@@ -31,74 +30,51 @@ func GetAllApplicationsForRepo(ctx context.Context, repository *models.Repositor
...
 
-	result, err := DefaultDeployer.DeployAndWait(ctx, &job.Application, content)
-	if err != nil {
-		log.Error().Err(err).Str("applicationId", job.Application.Id).Msg("failed to deploy compose file to agent")
-		return
-	}
-	if result == nil {
-		log.Error().Str("applicationId", job.Application.Id).Msg("agent deployment finished without a result")
-		return
+	if err := deployer.TriggerApplicationDeploy(context.Background(), &job.Application, content); err != nil {
```
alex289, line 80: **Claude Code Review: 🔴 Critical — compose file / commit never persisted on success**

`TriggerApplicationDeploy` is fire-and-forget: it dispatches the message and returns. But `ComposeFile`, `PreviousComposeFile`, `Commit`, and `CommitMessage` are never written to the DB in the changed-compose path.

The old `markDeploymentSuccess` carried an `updateFn` that did this atomically:
```go
update.ComposeFile = crypto.EncryptedString(content)
update.PreviousComposeFile = job.Application.ComposeFile
update.Commit = job.Commit
update.CommitMessage = job.CommitMessage
```
`handleDeployResult` in the websocket layer only updates `SyncStatus`, `HealthStatus`, `LastSyncedAt`, and `LastSyncError` — it has no access to `content`, `job.Commit`, or `job.CommitMessage`.

Result: after every compose-changing deployment, the DB record permanently shows the pre-deploy compose and commit. The application list will report stale data.

Fix: persist those fields optimistically before dispatching (or thread them through to the result handler).

backend/internal/hub/websocket/deploy.go
alex289: **Claude Code Review: 🔴 Bug — notification message shows UUID instead of app name**

```go
notifications.SendNotification(result.ApplicationId, "Success: deployment succeeded for "+result.ApplicationId, log)
```

`result.ApplicationId` is a UUID. The old code used `app.Name.String()`, so users saw "deployment succeeded for billing-service". Now they see "deployment succeeded for a3f2c1d8-...". Same issue on the failure path below.

backend/internal/hub/websocket/hub.go
```diff
@@ -63,7 +65,6 @@ func (h *Hub) Unregister(id string) {
 	h.mu.Lock()
 	delete(h.clients, id)
```
alex289, line 66: **Claude Code Review: 🔴 Bug — in-progress deploys stuck in `Syncing` after agent disconnect**

The old `Unregister` called `h.deployManager.FailPendingDeploys(id, ErrAgentDisconnected)`, which resolved any waiting deploy handle with an error so the caller could mark the app `OutOfSync`. That call was removed along with `DeployManager`.

Now, when an agent disconnects mid-deploy, the app stays in `Syncing` indefinitely. The disconnect handler sets `health_status = UnknownHealth` but leaves `sync_status = Syncing` — the user sees a permanently spinning deployment with no recovery path.

backend/internal/hub/websocket/handler.go
alex289: **Claude Code Review: 🟡 Concern — bulk-setting all apps `Healthy` on agent connect may mask real failure state**

An application that was `OutOfSync + Unhealthy` due to a failed deploy will be silently reset to `Healthy` the instant the agent reconnects, with no verification that the containers are actually running.

Consider using `UnknownHealth` here instead — the agent reconnecting tells us the agent is reachable, not that the containers are healthy.

backend/internal/hub/routes/applications.go
alex289: **Claude Code Review: 🟢 Style — redundant `else` after `return`**

```go
if err != nil {
    c.JSON(http.StatusInternalServerError, ...)
    return
} else {               // ← unnecessary; golangci-lint will flag this
    c.JSON(http.StatusAccepted, ...)
}
```

Drop the `else` — the `if` already returns.

backend/internal/hub/routes/applications.go
```diff
@@ -131,6 +135,27 @@ func GetApplicationHandler(c *gin.Context) {
 	c.JSON(http.StatusOK, toApplicationResponse(&application))
 }
 
+// applicationNameTaken reports whether another application already uses name
+// (case-insensitive). Names are stored encrypted with a random nonce, so a DB
+// unique index can't enforce this — we decrypt and compare in code. excludeID
+// skips a record (the one being updated).
+func applicationNameTaken(ctx context.Context, name, excludeID string) (bool, error) {
+	apps, err := gorm.G[models.Application](db.DB).Find(ctx)
```
alex289: **Claude Code Review: 🟢 Minor — `applicationNameTaken` is O(n) with decryption per row**

This loads every application from the DB, then decrypts and compares each name individually. Since names use random-nonce encryption a DB unique index is not possible (as the comment explains), but the full-table scan + per-row decrypt will degrade linearly with application count. Worth a comment flagging the limitation so future maintainers are not surprised.

backend/internal/hub/applications/repo_sync.go
```diff
@@ -23,43 +41,43 @@ func SyncRepository(ctx context.Context, repo *models.Repository, log *zerolog.L
...
+		hash, message, err := resolve(ctx, branch)
 		if err != nil {
-			lastErrMsg = fmt.Sprintf("failed to get latest commit for branch %q: %v", branch, err)
-			log.Error().Err(err).Str("repositoryId", repo.Id).Str("branch", branch).Msg("failed to get latest commit during sync")
+			lastErrMsg = fmt.Sprintf("failed to resolve commit for branch %q: %v", branch, err)
+			log.Error().Err(err).Str("repositoryId", repo.Id).Str("branch", branch).Msg("failed to resolve commit during sync")
 			continue
 		}
-		DefaultQueue.Enqueue(repo, provider, branchApps, commitInfo.Hash, commitInfo.Message)
+		if DefaultQueue != nil {
```
alex289, line 78: **Claude Code Review: 🟢 Minor — silent no-op when `DefaultQueue` is nil**

If `DefaultQueue` is nil (e.g. a misconfigured startup), applications are silently skipped without any log entry. A misconfiguration in production would look like a quiet sync with no errors. Consider adding a `log.Error()` or returning an error so the caller can surface it.

alex289 (commented):
**Claude Code Review** — Full findings below as inline comments. Overall direction is solid; the main concern is a data-loss bug in the fire-and-forget deploy path.

alex289:
Codex Review:

I found a few behavioral regressions that are worth fixing before merge:

1. P1 - Changed-compose syncs no longer persist the new manifest or commit metadata.

`backend/internal/hub/applications/sync.go:57` dispatches the deploy with the fetched compose content, but the eventual success handler in `backend/internal/hub/websocket/deploy.go:24` only updates sync/health status, `LastSyncedAt`, and `LastSyncError`. It does not write `ComposeFile`, `PreviousComposeFile`, `Commit`, or `CommitMessage`. After a successful sync deploy, the app can still show the old manifest/commit and future syncs can keep seeing the same compose diff.

2. P1 - No-op syncs do not mark the application as synced anymore.

`backend/internal/hub/applications/sync.go:43` only updates `Commit` and `CommitMessage` when the compose file is unchanged. It no longer sets `SyncStatus=Synced`, `LastSyncedAt`, or clears `LastSyncError`, despite the comment saying the app is marked synced. An app that was previously `out_of_sync` can remain visibly failed after a successful no-op sync.

3. P1 - Delete can miss renamed applications.

`backend/internal/hub/routes/applications.go:394` sends the current application name to the agent, and the agent derives the Compose project/directory from that name in `backend/internal/agent/docker/deploy.go:172`. If the application was renamed after deployment, the stored name may no longer match the project name used by the running containers. Delete can then report success for the new name while leaving the old containers/directories behind.

4. P2 - Repository sync success does not clear stale errors.

`backend/internal/hub/applications/repo_sync.go:98` sets `LastSyncError: nil` via a struct update. GORM skips zero values in struct updates, so a repository can move back to `success` while retaining an old `last_sync_error`. Use `Select`, `Update`, or a map update when clearing nullable fields.

Verification: `go test ./internal/hub/applications ./internal/hub/websocket ./internal/hub/deployer ./internal/hub/routes` passes, so these look like coverage gaps rather than failing-test regressions.

backend/internal/agent/docker/deploy.go
maximbigler: Created issue for this :D

frontend/src/routes/_authenticated/applications/$id/details.tsx
```diff
@@ -197,6 +207,23 @@ function ApplicationDetailsPage() {
 
 	return (
 		<div className="space-y-6">
+			<Breadcrumb>
```
alex289: Image

What is this?

backend/internal/hub/server.go
```diff
@@ -128,6 +129,15 @@ func Run(cfg Config) error {
 		Log.Warn().Err(resetSyncStatus.Error).Msg("failed to reset repositories stuck in syncing status")
 	}
 
+	// Populate the application name_hash blind index for rows created before the
+	// column existed. Skipped in demo mode where the database is read-only.
+	if !cfg.Demo {
```
alex289: This should be in the db.go file where the migrations are applied and the demo data is filled int

backend/internal/agent/docker/status.go
```diff
@@ -0,0 +1,60 @@
...
+// an application, identified by the orca-cd.application-id label applied at
+// deploy time. It returns HealthUnknown when the daemon is unreachable or no
+// containers exist (nothing deployed yet). Any container that is not running, or
+// whose healthcheck reports unhealthy, makes the whole application unhealthy.
+func (c *Client) ApplicationHealth(ctx context.Context, appID string) HealthState {
+	if !c.Ready() {
+		return HealthUnknown
+	}
+
+	result, err := c.cli.Client().ContainerList(ctx, client.ContainerListOptions{
```
alex289, line 30: This only checks if the container exists right? Not if it's running or if any configured health checks are healthy right?

alex289: Ah sorry. Just should have watched below

backend/internal/hub/crypto/crypto.go
```diff
@@ -15,7 +16,8 @@ import (
 var defaultCipher *Cipher
 
 type Cipher struct {
-	aead cipher.AEAD
+	aead    cipher.AEAD
+	hashKey []byte
```
alex289: This sounds like it's a general hashkey and not specific for the db blind index key

alex289:
Claude Review:

Apps can get stuck in Syncing forever — lost timeout + disconnect handling

The old path had two safety nets that are now gone:
- manualDeployTimeout (5 min) in TrackManualDeploy — marked the app failed if the agent never returned a result.
- FailPendingDeploys on Unregister — failed in-flight deploys when the agent disconnected.

Both were deleted. Now deploys are fire-and-forget and only resolve when a DeployResult arrives via handleDeployResult. If the agent crashes or disconnects mid-deploy, no result ever arrives, and I confirmed:
- On disconnect, handler.go only resets health_status → UnknownHealth; SyncStatus stays Syncing.
- server.go's boot-time resetSyncStatus resets repositories stuck in syncing — there is no equivalent for applications.

So a crashed agent (or a hub restart during a deploy) leaves applications pinned at Syncing indefinitely. Recommend either re-failing pending deploys on agent unregister, adding an app-level Syncing reset on boot, or a deploy-timeout sweeper.

alex289:
Manual deploy on an offline agent now returns 500 (was 409), and the test no longer covers it

DeployApplicationHandler (routes/applications.go:442) calls fire-and-forget TriggerApplicationDeploy, which does return an error when hub.Send fails — so the handler hits c.JSON(http.StatusInternalServerError, …). The old code returned 409 Conflict ("agent is not connected"), which is the correct semantic and is exactly what the new DeleteApplicationHandler still does (ErrAgentOffline → 409). So delete and deploy are now inconsistent for the same offline condition.

Worse, TestDeployApplicationHandler_AgentUnavailable was changed to use a bare &stubRouteDeployer{} (no startErr), so it returns nil and the test asserts 202 with a comment claiming "Agent-not-connected handling is still a TODO." The test no longer exercises the offline path at all — its name is now misleading. Either restore a stub that simulates the offline error and assert 409, or rename the test to reflect what it checks.

alex289:
Delete timeout mismatch: hub 30s vs. agent 5min → state divergence

agentDeleteTimeout = 30s (hub) but the agent's executeDelete uses deploymentTimeout = 5 * time.Minute for compose down. If teardown takes 30s–5min, the hub times out (504), keeps its record, and tells the user it failed — but the agent goes on to complete the removal (containers + deployment dir gone). The hub now references an application the agent has destroyed: the inverse of the "don't lose track of running containers" goal this flow is designed around. Consider aligning the hub wait with the agent's execution timeout, or making the agent abort teardown when its delete context is cancelled.

backend/internal/hub/deployer/application_deployer.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	if !d.hub.Send(app.AgentId, msg) {
+		errMsg := "could not connect to agent: " + app.Agent.Name.String()
+		_ = updateApplicationStatus(ctx, app.Id, models.Application{
+			SyncStatus:    models.OutOfSync,
+			HealthStatus:  models.UnknownHealth,
+			LastSyncError: &errMsg,
+		}, d.log)
+		return errors.New("Error: " + errMsg)
```
alex289:
```suggestion
		return errors.New("error: " + errMsg)
```

backend/internal/hub/deployer/application_deployer.go
```diff
@@ -0,0 +1,93 @@
...
+	return &ApplicationDeployer{
+		hub: hub,
+		log: log,
+	}
+}
+
+var DefaultApplicationDeployer ApplicationDeploymentManager
+
+func (d *ApplicationDeployer) TriggerApplicationDeploy(ctx context.Context, app *models.Application, composeFile string) error {
+
```
alex289, line 49:
```suggestion
```

frontend/src/routes/_authenticated/applications/$id/details.tsx
maximbigler: I think that are merge conflicts error. I will remove it :D

alex289 (approved):

backend/internal/hub/db/migrations/000023_add_application_name_hash.up.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE applications ADD COLUMN name_hash TEXT NOT NULL DEFAULT '';
+
+-- Partial index excludes legacy rows (empty hash, pre-backfill) so they don't
+-- collide on (agent_id, ''). Names were globally unique under the old rule, so
```
timokoessler: We did not check this before no?

backend/internal/hub/routes/applications.go
```diff
@@ -131,6 +138,22 @@ func GetApplicationHandler(c *gin.Context) {
 	c.JSON(http.StatusOK, toApplicationResponse(&application))
 }
 
+// applicationNameTaken reports whether another application on the same agent
+// already uses name (case-insensitive). Names are stored encrypted with a random
+// nonce, so we match on the name_hash blind index instead of decrypting every
+// row. excludeID skips a record (the one being updated). Uniqueness is scoped per
+// agent — the same name may exist on different agents.
+func applicationNameTaken(ctx context.Context, name, agentID, excludeID string) (bool, error) {
```
timokoessler: In theory we could just run the insert and catch duplicate key error like done elsewhere?

alex289: Good idea

alex289 (approved):

PR #215, opened 2026-07-06
feat: Add --help and --version to root command

timokoessler:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

The `--help` option only worked for sub-commands before.

### DIFF
--- a/backend/cmd/agent/main.go
+++ b/backend/cmd/agent/main.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"slices"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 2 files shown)


backend/cmd/agent/main.go
```diff
@@ -24,6 +25,14 @@ func main() {
 		DisableFlagParsing: true,
 		SilenceUsage:       true,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if slices.Contains(args, "--help") || slices.Contains(args, "-h") {
```
alex289, line 28: Aren't the arguments with a single - always uppercase?

alex289 (approved):

backend/cmd/agent/main.go
timokoessler: Actually many Linux CLI tools have lowercase and uppercase short options with different meanings, but I think in Linux world lower case is the default.

timokoessler: Old Windows cmd tools always use uppercase.

PR #217, opened 2026-07-06
feat: Add env var to limit allowed ip addresses

timokoessler:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

Adds a `ALLOWED_IPS` env var that allows locking most endpoints down to trusted ips. Excludes Webhooks, healthcheck and agent websocket.


backend/internal/hub/middleware/iplock.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/OrcaCD/orca-cd/internal/shared/iplist"
+	"github.com/gin-gonic/gin"
+)
+
+var ipLockExemptPrefixes = []string{
+	"/api/v1/webhooks/",
+	"/api/v1/github-actions",
+}
+
+var ipLockExemptPaths = map[string]bool{
```
alex289, line 16: Why isn't this also just a string array?

alex289 (approved):

backend/internal/hub/middleware/iplock.go
timokoessler: It's faster, as this behaves like a JS Set which sadly does not exist in Go.

PR #230, opened 2026-07-12
feat: Application event history

### LINKED ISSUE #228: feat: Application event history
When looking at the application details there is no list of events.

It would be useful to see a list of events with a timestamp, the trigger (commit hash or image tag) which lead to the update or even if nothing was updated it would also be useful

alex289:
Fixes #228

Bildschirmfoto 2026-07-12 um 19 41 22

## Summary

Adds a live, application-scoped history of deployments, commit synchronizations, and image updates with source/actor attribution, terminal results, and bounded retention.

- **Schema**: new `application_events` table (migration 24) with `ON DELETE CASCADE` for applications and `ON DELETE SET NULL` for actor users
- **Lifecycle service** (`applicationevents`): start/complete correlation via shared protobuf request IDs, terminal records, per-application retention capped at 1,000 events, and restart recovery
- **API**: protected `GET /api/v1/applications/:id/events` with offset pagination (`{items, hasMore}`)
- **Recording**: manual/initial deployments, commit syncs (manual, polling, webhook, GitHub Actions), and image updates (webhook, GitHub Actions, periodic polling)
- **Noise control**: scheduled polls that resolve to the recorded commit or successful no-op image polls leave no history; explicit triggers always record, including `no_change`
- **Recovery**: hub startup marks interrupted `running` events as failed
- **Frontend**: new `/applications/$id/history` page with SWR Infinite pagination, SSE live updates, localized table (en/de), expandable failure details, and sidebar/breadcrumb integration

History persistence failures are logged and never block the underlying deployment, sync, or image operation.


backend/internal/hub/applications/sync.go
```diff
@@ -32,11 +42,24 @@ func GetAllApplicationsForRepo(ctx context.Context, repository *models.Repositor
 }
 
 func processSyncJob(ctx context.Context, job syncJob, log *zerolog.Logger) {
+	// Scheduled polls that resolve to the already-recorded commit are pure no-ops:
+	// skip them entirely so the history is not flooded every polling interval.
+	// Explicit triggers (manual, webhook, CI) always leave a history record.
+	if job.Origin.Source == models.ApplicationEventSourceRepositoryPolling && job.Commit == job.Application.Commit {
+		return
```
timokoessler, line 48: This does not only skip the event recording, but also the other steps afterwards, so the real logic?

backend/internal/hub/db/migrations/000024_create_application_events_table.up.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE IF NOT EXISTS application_events (
+    id             TEXT PRIMARY KEY,
+    application_id TEXT NOT NULL,
+    request_id     TEXT,
+    type           TEXT NOT NULL CHECK (type IN ('deployment', 'commit_sync', 'image_update')),
```
timokoessler: Nice to add checks, but we don't do it for any other table, so maybe remove and keep consistently checking on the application layer?

backend/internal/hub/applicationevents/events.go
```diff
@@ -0,0 +1,141 @@
...
+	status models.ApplicationEventStatus,
+	errorMessage *string,
+) (*models.ApplicationEvent, error) {
+	if !isTerminal(status) {
+		return nil, fmt.Errorf("terminal event cannot have running status")
+	}
+	return create(ctx, p, status, errorMessage)
+}
+
+func isTerminal(status models.ApplicationEventStatus) bool {
```
timokoessler: I don't really get what the terminal thing is about from reading this file.

backend/internal/hub/applicationevents/events.go
```diff
@@ -0,0 +1,141 @@
...
+	if status != models.ApplicationEventRunning {
+		now := time.Now()
+		event.CompletedAt = &now
+	}
+
+	err := db.DB.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		if err := gorm.G[models.ApplicationEvent](tx).Create(ctx, &event); err != nil {
+			return err
+		}
+		return tx.Exec(`DELETE FROM application_events
```
timokoessler: Move to separate method so it's clear that this is cleaning up old events after limit? Also not sure if this needs to be a transaction?

backend/internal/hub/websocket/deploy.go
```diff
@@ -34,8 +35,10 @@ func handleDeployResult(result *messages.DeployResult, log *zerolog.Logger) {
 			LastSyncError: &cleared,
 		}, log)
 		if err != nil {
+			completeDeployEvent(ctx, result, models.ApplicationEventSucceeded, nil, log)
```
timokoessler: success if err != nil?

backend/internal/hub/websocket/image_poll.go
```diff
@@ -59,3 +62,44 @@ func handlePullImagesResult(client *Client, r *messages.PullImagesResult, log *z
 
 	sse.PublishUpdate("/api/v1/applications")
 }
+
+// recordImagePullHistory completes the explicit image_update event matching this
+// result, or persists a terminal image_polling event for unsolicited periodic
```
timokoessler: "persists a terminal image_polling" interesting way to tell that it logs something 😅

frontend/src/routes/_authenticated/applications/$id/history.tsx
```diff
@@ -0,0 +1,85 @@
...
+		if (previousPageData && !previousPageData.hasMore) {
+			return null;
+		}
+		const offset = pageIndex * APPLICATION_EVENTS_LIMIT;
+		return `${API_BASE}/applications/${id}/events?limit=${APPLICATION_EVENTS_LIMIT}&offset=${offset}`;
+	};
+
+// Keys include API_BASE so the SSE broker's update events (which publish full
+// API paths) revalidate every loaded page of this list.
+async function fetchEventsPage(url: string): Promise<ApplicationEventsPage> {
```
timokoessler: Why not our existing fetcher?

timokoessler (approved):

backend/internal/hub/applications/sync.go
```diff
@@ -42,14 +42,15 @@ func GetAllApplicationsForRepo(ctx context.Context, repository *models.Repositor
...
-	// A scheduled poll for the already-recorded commit still performs the sync so
-	// application status is refreshed, but it does not add a repetitive event.
-	if job.Origin.Source != models.ApplicationEventSourceRepositoryPolling || job.Commit != job.Application.Commit {
-		requestID = uuid.NewString()
-		if _, err := applicationevents.Start(ctx, syncEventParams(job, &requestID)); err != nil {
-			log.Error().Err(err).Str("applicationId", job.Application.Id).Msg("failed to record commit sync event")
-		}
+	// Scheduled polls are edge-triggered by commit. Reprocessing an already-recorded
+	// commit could incorrectly clear a prior deployment failure without redeploying.
+	if job.Origin.Source == models.ApplicationEventSourceRepositoryPolling && job.Commit == job.Application.Commit {
```
timokoessler, line 47: I don't really get why we only have a early return for polling here? And why it's there at all? The correct way would be to let if continue and then it notices that compose file did not change? I think I already saw a guard against a full history log of polling events somewhere else in the logic?

alex289: The early return and the later compose comparison cover two different cases. `job.Commit == job.Application.Commit` means polling found no new Git state. Since commits are immutable, fetching the same compose file again provides no new information. If processing continued, we would create an event on every polling interval and then complete it as `no_change`, which is exactly the history noise this guard is intended to prevent. More importantly, the unchanged-compose branch marks the application synced and clears `LastSyncError`; doing that for an already-recorded commit could incorrectly erase a previous deployment failure without redeploying anything.

The later `content == job.Application.ComposeFile.String()` check handles the meaningful case: a new commit was found, but that commit did not change this application compose file. That case should update the recorded commit and create a `no_change` event.

It only returns early for polling because manual, webhook, and GitHub Actions syncs are explicit operations that should remain visible even when repeated. The similar guard in `recordSyncFailure` only prevents pre-job queue/resolver failures from creating history for the same unchanged polling commit; the poller guards themselves only control scheduling and concurrent repository syncs.

So the intended behavior is: polling + same commit → skip; polling + new commit + same compose → `no_change`; polling + new commit + changed compose → deploy; explicit sync + same commit → record the explicit `no_change` operation.

alex289: @timokoessler please resolve or comment :)

timokoessler (approved):

PR #232, opened 2026-07-13
feat: Security policy to detect privileged containers

timokoessler:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

Try to detect common dangerous pattern allowing a container to access the host system.

### DIFF
--- a/backend/internal/agent/docker/policy.go
+++ b/backend/internal/agent/docker/policy.go
@@ -1,0 +1,212 @@
+package docker
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	composetypes "github.com/compose-spec/compose-go/v2/types"
+)
+
+// Linux capabilities that grant meaningful host access when added via cap_add.
+var dangerousCapabilities = map[string]struct{}{
+	"ALL":             {},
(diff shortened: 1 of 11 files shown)


backend/internal/agent/docker/policy.go
```diff
@@ -0,0 +1,221 @@
...
+
+// dangerous security_opt values
+var unsafeSecurityOpts = map[string]struct{}{
+	"seccomp:unconfined":  {},
+	"apparmor:unconfined": {},
+}
+
+// Top-level directories under which sensitivePathComponents is checked.
+var sensitiveComponentRoots = []string{
+	"/home",
```
alex289, line 57: Does this mean I can't use /home/... as a volume? We currently can only use absolute paths in the volume so this will break this right?

alex289 (approved):

alex289:
Don't forget to add new env variables to docs

backend/internal/agent/docker/policy.go
timokoessler: You can use /home/something, but it will check against the list of unsafe dirs, like /home/user/.ssh.

PR #234, opened 2026-07-14
fix: Support relative bind mounts from agent containers

### LINKED ISSUE #231: bug: Absolute path required for volumes
Applications with volumes require an absolute path to work somehow. This is annoying

alex289:
## **Type of change**

- [x] 🐛 Bug fix
- [ ] 🚀 New feature

## **Description**

Fixes #231


backend/internal/agent/docker/docker.go
```diff
@@ -85,6 +96,57 @@ func (c *Client) Ready() bool {
...
+func (c *Client) hostDeploymentsBase() string {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.hostDeploymentsDir
+}
+
+func (c *Client) resolveHostDeploymentsDir(ctx context.Context) {
+	c.mu.Lock()
+	if c.hostDeploymentsDir != "" {
+		c.mu.Unlock()
```
timokoessler: defer Unlock once?

backend/internal/agent/docker/docker.go
```diff
@@ -85,6 +96,57 @@ func (c *Client) Ready() bool {
...
+	if detector == nil {
+		err = fmt.Errorf("host deployments directory detector is not configured")
+	} else {
+		detectionCtx, cancel := context.WithTimeout(ctx, hostDirDetectionTimeout)
+		hostDeploymentsDir, err = detector(detectionCtx)
+		cancel()
+	}
+
+	c.mu.Lock()
+	if err == nil {
```
timokoessler, line 186: Not sure if early return makes this better readable?

backend/internal/agent/docker/mounts.go
```diff
@@ -0,0 +1,166 @@
...
+	ContainerList(context.Context, client.ContainerListOptions) (client.ContainerListResult, error)
+}
+
+func detectHostDeploymentsDir(
+	ctx context.Context,
+	inspector containerInspector,
+	hostname func() (string, error),
+	deploymentsDir string,
+) (string, error) {
+	containerID, err := hostname()
```
timokoessler: Claude says it's safer tot parse /proc/self/mountinfo or /proc/self/cgroup

backend/internal/agent/docker/mounts.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/moby/moby/api/types/mount"
+	"github.com/moby/moby/client"
+)
+
+type containerInspector interface {
+	ContainerInspect(context.Context, string, client.ContainerInspectOptions) (client.ContainerInspectResult, error)
+	ContainerList(context.Context, client.ContainerListOptions) (client.ContainerListResult, error)
+}
+
+func detectHostDeploymentsDir(
```
timokoessler: Does this work gracefully in dev outside of container?

backend/internal/agent/docker/mounts.go
timokoessler: (User can define container hostname)

backend/internal/agent/docker/docker.go
```diff
@@ -85,6 +99,94 @@ func (c *Client) Ready() bool {
 	return c.ready
 }
 
+func (c *Client) hostDeploymentsBase() string {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.hostDeploymentsDir
+}
+
+func (c *Client) resolveHostDeploymentsDir(ctx context.Context) {
```
timokoessler: If a caller whose own ctx gets canceled while another goroutine's detection is still in flight returns immediately with an empty/stale hostDeploymentsDir, so that one Deploy/CheckAndPullImages call silently skips translation (falls back to pre-PR behavior for that single call). Self-heals on the next call once detection completes.

timokoessler: + What if docker agent container uses named volume

timokoessler (approved):

PR #245, opened 2026-07-16
fix: Deployment waiting on app being healthy

### LINKED ISSUE #110: feat: Health status analysis of applications

### LINKED ISSUE #240: bug: Deployment waiting on app being healthy
This might be partially intended, but results in the app's status being deploying / pending until the app is health. I think we should refactor to do the health checks after the deployment process.

alex289:
## **Type of change**

- [x] 🐛 Bug fix
- [ ] 🚀 New feature

## **Description**

**The cause**: the agent's compose up ran with `Start: {Wait: true, WaitTimeout: 2min}`, which blocked until every container passed its healthcheck. The DeployResult (and therefore the app's sync status) was held hostage by health, so the app stayed in "deploying" until healthy — or became "failed" after two minutes if a healthcheck never passed.

Changes:

- Agent deploy (`agent/docker/deploy.go`): removed the wait option so compose up returns once containers are started and the DeployResult is sent immediately. The image-update path received the same fix.
- Event-driven health (`agent/docker/status.go`, `agent/docker/health_watcher.go`): the agent derives aggregate application health from Docker container lifecycle and health-status events. Event bursts are debounced, with a hard maximum delay so crash loops cannot postpone reporting indefinitely. A healthcheck in the starting state remains unknown until it settles.
- Recovery and reconciliation (`agent/docker/docker.go`): health is resynchronized on initial daemon connection, after reconnects, and periodically. Each successful snapshot also removes watcher state for applications whose containers disappeared outside OrcaCD, such as after a manual compose down.
- Hub (`hub/websocket/deploy.go`): a successful deployment now sets only the sync status to synced. Health stays unknown until the agent reports the observed container state.

Net behavior: an app becomes synced as soon as its containers are up, shows unknown health while healthchecks settle, and then changes to healthy or unhealthy from the agent's report. Later health transitions are monitored continuously. A slow or broken healthcheck can no longer fail or stall the deployment itself.

## **Additional context**

Fixes #110
Fixes #240


alex289:
@timokoessler does this also fix #110?

backend/internal/agent/websocket.go
```diff
@@ -285,6 +278,31 @@ func executeDeployment(ctx context.Context, sender outboundSender, deployer depl
...
+	// health now that its containers are up and report it once it settles.
+	if waiter, ok := deployer.(applicationHealthWaiter); ok {
+		reportPostDeployHealth(ctx, sender, waiter, req.ApplicationId)
+	}
+}
+
+// reportPostDeployHealth waits for the freshly deployed application's health to
+// settle (bounded by docker.HealthSettleTimeout) and sends the outcome to the
+// hub as an ApplicationStatusReport.
+func reportPostDeployHealth(ctx context.Context, sender outboundSender, waiter applicationHealthWaiter, appID string) {
```
timokoessler: I think we won't need this in the future if we can observe health changes via the docker deamon events we are already subscribing to? Not sure if this would be easy or complicated to implement tho.

timokoessler (approved):

timokoessler:
> @timokoessler does this also fix #110?

No, as we don't monitor for any health changes after the deployment (see PR comment above)

timokoessler:
See this code [here](https://github.com/OrcaCD/orca-cd/blob/56d956a702169b4baf40357f28187ae545b0f720/backend/internal/agent/docker/docker.go#L257), I never implemented it :D

backend/internal/agent/websocket.go
alex289: Done. And now it should resolve #110 right? :D

timokoessler:
The AI generated PR description is now outdated 😛

backend/internal/agent/docker/health_watcher.go
```diff
@@ -0,0 +1,191 @@
...
+
+// healthRelevantAction reports whether a container event action can change the
+// aggregate health derived in aggregateHealth (running state or healthcheck
+// status). Notably this excludes the exec_* events emitted for every
+// healthcheck run.
+func healthRelevantAction(action events.Action) bool {
+	switch action {
+	case events.ActionStart, events.ActionRestart, events.ActionStop, events.ActionDie,
+		events.ActionOOM, events.ActionPause, events.ActionUnPause, events.ActionDestroy,
+		events.ActionHealthStatusRunning, events.ActionHealthStatusHealthy, events.ActionHealthStatusUnhealthy:
```
timokoessler: There is also a `events.ActionHealthStatus` event. Not sure whether it not send in addition to a specific event e.g. `ActionHealthStatusRunning` or in different cases.

backend/internal/agent/docker/status.go
```diff
@@ -42,19 +56,31 @@ func (c *Client) ApplicationHealth(ctx context.Context, appID string) HealthStat
...
 			return HealthUnhealthy
 		}
-		if item.Health != nil && item.Health.Status == container.Unhealthy {
+		if item.Health == nil {
+			continue
+		}
+		switch item.Health.Status {
+		case container.Unhealthy:
 			return HealthUnhealthy
+		case container.Starting:
```
timokoessler, line 77: In the future we should maybe support starting as a health state in the hub? Maybe we should create a issue or add a todo comment if you agree

backend/internal/agent/docker/health_watcher.go
```diff
@@ -0,0 +1,191 @@
...
+func (c *Client) observeApplicationHealth(appID string) {
+	c.healthWatcher.invalidate(appID)
+	c.healthWatcher.schedule(appID)
+}
+
+// healthRelevantAction reports whether a container event action can change the
+// aggregate health derived in aggregateHealth (running state or healthcheck
+// status). Notably this excludes the exec_* events emitted for every
+// healthcheck run.
+func healthRelevantAction(action events.Action) bool {
```
timokoessler: nitpick: add `is` prefx as it returns a boolean?

backend/internal/agent/docker/docker.go
```diff
@@ -254,8 +257,41 @@ func (c *Client) startEventLoop() {
...
+		return
+	}
+	appID := msg.Actor.Attributes[labelApplicationID]
+	if appID == "" || !healthRelevantAction(msg.Action) {
+		return
+	}
+	c.healthWatcher.schedule(appID)
+}
+
+// resyncApplicationHealth re-evaluates the health of every known application.
```
timokoessler: This is only called in waitForDaemon(), but not triggered on agent startup in case the connection to Docker deamon is successfully in first try.

backend/internal/agent/docker/health_watcher.go
```diff
@@ -0,0 +1,191 @@
...
+// schedule queues a debounced health evaluation for appID, extending the
+// window if one is already pending.
+func (w *healthWatcher) schedule(appID string) {
+	if w == nil {
+		return
+	}
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	if timer, ok := w.pending[appID]; ok {
+		timer.Reset(w.debounce)
```
timokoessler: In case an app is always in a crashing loop, we will never report the status to the hub. I would suggest that we always send the health status event after x events or after a amount of time?

backend/internal/agent/docker/docker.go
```diff
@@ -254,8 +257,41 @@ func (c *Client) startEventLoop() {
...
+// Called after the event stream was interrupted, since health transitions may
+// have been missed while the daemon was unreachable.
+func (c *Client) resyncApplicationHealth() {
```
timokoessler, line 276: Maybe we should call this function periodically and clean up known apps instead of appending? There are some cases right now there an application will not be removed from knowApps, e.g. if the user does docker compose down.

timokoessler (commented):
Nice PR!

timokoessler (approved):

timokoessler (approved):

backend/internal/hub/routes/agents.go
```diff
@@ -319,8 +319,7 @@ func RotateAgentTokenHandler(c *gin.Context) {
 	utils.RecordAuditLog(c, "rotated-token", "agent", agent.Id)
 
 	if client, ok := websocket.DefaultHub.GetClient(agent.Id); ok {
-		websocket.DefaultHub.Unregister(agent.Id)
-		client.Close()
+		websocket.DefaultHub.BeginDisconnect(client)
```
timokoessler: This change looks a bit dangerous to me, as BeginDisconnect has a early return, and we always need to disconnect the client in case it is still connected? I also would expect it to happen immediately?

backend/internal/hub/websocket/image_poll.go
```diff
@@ -54,13 +53,13 @@ func handlePullImagesResult(client *Client, r *messages.PullImagesResult, log *z
 
 	recordImagePullHistory(ctx, r, log)
 
+	sse.PublishUpdate("/api/v1/applications")
+
 	if r.Success {
-		notifications.SendNotification(r.ApplicationId, "Success: image update succeeded for "+app.Name.String(), log)
+		dispatchNotification(r.ApplicationId, "Success: image update succeeded for "+app.Name.String(), log)
```
timokoessler: That basically runs it in a go thread or? Why can't we simply write a go in front of it?

backend/internal/hub/websocket/handler.go
```diff
@@ -126,8 +130,10 @@ func WsHandler(h *Hub, log *zerolog.Logger) gin.HandlerFunc {
 		}
 
 		defer func() {
-			h.Unregister(claims.Subject)
-			client.Close()
+			// Reserve the registration until cleanup finishes. Otherwise a new
+			// session can report fresh state that this old session then overwrites.
+			h.BeginDisconnect(client)
+			defer h.FinishDisconnect(client)
```
timokoessler, line 145: We are calling `FinishDisconnect` but I think not in case of token revocation?

backend/internal/hub/websocket/handler.go
```diff
@@ -183,15 +201,29 @@ func handleClientMessage(client *Client, msg *messages.ClientMessage, log *zerol
...
 			log.Warn().Str("client", client.Id).Msg("Dropping doubly-encrypted message")
-			return
+			return nil, false
 		}
 		msg = inner
 	}
+	return msg, true
+}
+
+func dispatchClientMessage(msg *messages.ClientMessage, handle func()) {
```
timokoessler: Why do we only want to handle some client messages in a go routine? Fells much safer to do this for all us until now to not block connection?

backend/internal/hub/websocket/handler.go
```diff
@@ -165,16 +171,28 @@ func WsHandler(h *Hub, log *zerolog.Logger) gin.HandlerFunc {
 				continue
 			}
 
-			go handleClientMessage(client, msg, log)
+			decoded, ok := decodeClientMessage(client, msg, log)
+			if !ok {
+				continue
+			}
+			dispatchClientMessage(decoded, func() { handleDecodedClientMessage(client, decoded, log) })
```
timokoessler: Why this function argument if always the same function is called? Makes it more complex? For testing?

backend/internal/agent/docker/health_watcher.go
```diff
@@ -0,0 +1,498 @@
...
+	w.mu.Unlock()
+	w.reportMu.Unlock()
+
+	for evaluation := range evaluations {
+		<-evaluation.done
+	}
+}
+
+func (w *healthWatcher) resetLocked(appID string) *healthEvaluation {
+	delete(w.lastReported, appID)
```
timokoessler, line 216: No mutex needed here?

backend/internal/hub/websocket/image_poll.go
```diff
@@ -54,13 +57,18 @@ func handlePullImagesResult(client *Client, r *messages.PullImagesResult, log *z
...
 	if r.Success {
-		notifications.SendNotification(r.ApplicationId, "Success: image update succeeded for "+app.Name.String(), log)
-	} else {
-		notifications.SendNotification(r.ApplicationId, "Error: image update failed for "+app.Name.String(), log)
+		return &notificationRequest{
+			applicationID: r.ApplicationId,
+			message:       "Success: image update succeeded for " + app.Name.String(),
+		}
+	}
+	return &notificationRequest{
```
timokoessler: I still don't get the reason for this change?

backend/internal/hub/websocket/handler.go
```diff
@@ -221,14 +325,21 @@ func handleClientMessage(client *Client, msg *messages.ClientMessage, log *zerol
 // applications. Updates are scoped to the reporting agent so an agent can only
 // affect its own applications.
 func handleApplicationStatusReport(client *Client, report *messages.ApplicationStatusReport, log *zerolog.Logger) {
```
timokoessler: I think this function is not called anywhere else, so why create a new one with a context arg instead of modifying this one without renaming? Then we can also get rid of the bad context.Background()?

backend/internal/hub/websocket/handler.go
```diff
@@ -183,35 +246,76 @@ func handleClientMessage(client *Client, msg *messages.ClientMessage, log *zerol
...
+func queueClientMessage(messageQueue chan<- *messages.ClientMessage, msg *messages.ClientMessage) bool {
+	select {
+	case messageQueue <- msg:
+		return true
+	default:
+		return false
+	}
+}
+
+func handleClientMessages(ctx context.Context, client *Client, messageQueue <-chan *messages.ClientMessage, log *zerolog.Logger) {
```
timokoessler: We now have a lot of handleClientMessage functions, although the existing one was only used once, so why create new ones?

backend/internal/hub/websocket/deploy.go
```diff
@@ -7,56 +7,72 @@ import (
...
 )
 
-func handleDeployResult(result *messages.DeployResult, log *zerolog.Logger) {
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+type notificationRequest struct {
+	applicationID string
+	message       string
+}
+
+func handleDeployResult(result *messages.DeployResult, log *zerolog.Logger) *notificationRequest {
```
timokoessler: Can't we get rid of the method without context and do not rename the method? It's only called once?

timokoessler (approved):

PR #250, opened 2026-07-19
fix: Loading OIDC profile pictures

alex289:
## **Type of change**

- [x] 🐛 Bug fix
- [ ] 🚀 New feature

## **Description**

```
Loading the image 'https://oidc/api/users/userId/profile-picture.png' violates the following Content Security Policy directive: "img-src 'self' data:". The action has been blocked.
```

This fix proxies the image. Since the providers are dynamically configureable it was the only way possible instead of making the CSP softer


backend/internal/hub/routes/profile_picture.go
```diff
@@ -0,0 +1,70 @@
...
+
+	contentType := resp.Header.Get("Content-Type")
+	if resp.StatusCode != http.StatusOK || !strings.HasPrefix(strings.ToLower(contentType), "image/") {
+		c.JSON(http.StatusBadGateway, gin.H{"error": "failed to load profile picture"})
+		return
+	}
+
+	data, err := io.ReadAll(io.LimitReader(resp.Body, maxProfilePictureBytes+1))
+	if err != nil || len(data) > maxProfilePictureBytes {
+		c.JSON(http.StatusBadGateway, gin.H{"error": "failed to load profile picture"})
```
timokoessler: Maybe the error responses should contain more information to make debugging more easy?
Would also be cool to stream the response instead of loading into the RAM, but that sadly makes a nice error if the image is too large impossible.

timokoessler (approved):
Good find!

backend/internal/hub/routes/profile_picture.go
alex289: Added some debug logs

backend/internal/hub/routes/profile_picture.go
```diff
@@ -17,6 +18,8 @@ const profilePicturePath = "/api/v1/auth/profile-picture"
 
 const maxProfilePictureBytes = 5 << 20 // 5 MiB
 
+var profilePictureLog = logger.New("profile-picture", false)
```
timokoessler: This logger.New pattern breaks the log json setting (and maybe also log level?).
Maybe just return different error messages in HTTP response?

timokoessler (approved):

PR #259, opened 2026-07-22
fix: Agent and sync bugs

alex289:
## **Type of change**

- [x] 🐛 Bug fix
- [ ] 🚀 New feature

## **Description**

Fixes applications getting permanently stuck on "Syncing"/"Unknown" status and history filling up with spurious "Failed" (`hub restarted before the operation result was received`) entries, caused by duplicate webhook deliveries (e.g. GitHub retrying a `package` event for GHCR pushes).

- **Image pull webhook**: the agent silently dropped its response to the hub whenever an image pull found nothing new to pull — even for explicitly triggered requests (webhook/manual/GitHub Actions), not just background polling ticks. When GitHub redelivers a webhook, the second delivery finds the image already pulled by the first and never reports back. The corresponding `application_event` then sits in `running` forever (only cleared on the next hub restart, which is where the "hub restarted..." failures come from), and the application's `sync_status` never gets reset out of `Syncing`.
- **Commit-sync / deployment webhooks**: the same class of duplicate-delivery problem existed here too, plus a more serious gap — there was no locking at all preventing two overlapping deploys for the same application from running concurrently on the agent (`docker compose up` racing against itself for the same project directory).

## Notes for reviewers

- No commit/deploy correctness change is required beyond serialization — the deploy path already always reports a result back to the hub (unlike the image-poll path before this fix), so it doesn't get permanently "stuck"; it was purely a concurrency/race risk.
- The repository-level sync lock is released once the sync call returns (i.e. once jobs are enqueued), not once the deploy actually finishes. In practice this is fine because the agent-side lock added here serializes the actual work — a residual reordering (rather than corruption) risk remains for genuinely back-to-back distinct triggers, which is an acceptable tradeoff over the added complexity of holding the lock across async queue processing.

### DIFF
--- a/backend/internal/agent/websocket.go
+++ b/backend/internal/agent/websocket.go
@@ -252,6 +252,28 @@ func executePullImages(poller pollerHandler, req *messages.PullImagesRequest) {
 	poller.TriggerNow(req.ApplicationId, req.ApplicationName, req.RequestId)
 }
 
+// appExecLocks serializes Deploy/Remove calls per application ID. The hub has
+// no cross-request guarantee that only one deploy is ever in flight for a
+// given application (e.g. a duplicated webhook delivery can dispatch two
+// DeployRequests, or a delete can race a deploy), and executeDeployment /
+// executeDelete each run in their own goroutine. Without this lock, two
+// docker compose invocations for the same project could run concurrently.
+var (
+	appExecMu    sync.Mutex
+	appExecLocks = make(map[string]*sync.Mutex)
+)
+
+func lockForApplication(appID string) *sync.Mutex {
+	appExecMu.Lock()
+	defer appExecMu.Unlock()
+	lock, ok := appExecLocks[appID]
+	if !ok {
+		lock = &sync.Mutex{}
+		appExecLocks[appID] = lock
+	}
+	return lock
+}
+
 func executeDeployment(ctx context.Context, sender outboundSender, deployer deployExecutor, req *messages.DeployRequest) {
 	Log.Info().Str("application_id", req.ApplicationId).Str("request_id", req.RequestId).Msg("starting deployment")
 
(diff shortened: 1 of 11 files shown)


backend/internal/agent/websocket.go
```diff
@@ -252,6 +252,28 @@ func executePullImages(poller pollerHandler, req *messages.PullImagesRequest) {
...
 
+// appExecLocks serializes Deploy/Remove calls per application ID. The hub has
+// no cross-request guarantee that only one deploy is ever in flight for a
+// given application (e.g. a duplicated webhook delivery can dispatch two
+// DeployRequests, or a delete can race a deploy), and executeDeployment /
+// executeDelete each run in their own goroutine. Without this lock, two
+// docker compose invocations for the same project could run concurrently.
+var (
+	appExecMu    sync.Mutex
+	appExecLocks = make(map[string]*sync.Mutex)
```
timokoessler, line 263: Not sure if that important but we never clean appExecLocks up?

timokoessler (approved):
Cool. Do we still need additional deduplication for incoming webhooks in a short timeframe?

alex289:
> Cool. Do we still need additional deduplication for incoming webhooks in a short timeframe?

This should not be a problem I think. Even if both jobs end up in the queue the new agent lock serializes the docker compoes executions. So in worst case it will just be delayed

backend/internal/agent/websocket.go
alex289: Not a practical problem: entries only accumulate per distinct application ID ever deployed/deleted through this agent, not per deploy call — repeat deploys of the same app reuse the existing *sync.Mutex. For a self-hosted agent managing a handful to low-hundreds of apps over its lifetime, that's a handful to low-hundreds of ~48-byte mutex entries, even counting apps later removed from OrcaCD. It also mirrors the exact same never-cleaned-up pattern already in image_poller.go's runLocks map, so it's consistent with existing precedent in this codebase rather than a new risk.

timokoessler: A short answer would also be enough 😜

PR #264, opened 2026-07-25
feat: Image pull debounce

alex289:
## **Type of change**

- [ ] 🐛 Bug fix
- [x] 🚀 New feature

## **Description**

A single registry push often arrives as several webhook deliveries. GHCR publishes one
`package` event per manifest of a multi-arch image, so pushing one tag fired three image
update webhooks — three pull attempts, three history entries, and three notifications for
what the user experienced as one deployment.

Image update webhooks now go through a per-application debouncer on the hub. A delivery
queues the pull instead of dispatching it; further deliveries for the same application
restart a 5s quiet window, capped at 30s from the first delivery of the burst so a
continuous stream cannot postpone the pull indefinitely. Once the burst settles, exactly
one `PullImagesRequest` goes out, which means one history entry and one notification.

- `ImagePullDebouncer` (`internal/hub/applications/image_pull_debounce.go`) holds the
  pending pull per application, reloads the application before dispatching (it may have
  been changed or deleted while the burst settled), and drops queued pulls on hub shutdown.
- All three webhook flavours in `ImagePullWebhookHandler` (GitHub, Docker Hub,
  Harbor/generic) call `ScheduleImagePull`; the handler still responds `204` immediately.
- `TriggerImagePull` cancels a queued pull for the same application, so an immediate
  trigger is not followed by a redundant debounced one a few seconds later.
- The GitHub Actions endpoint keeps triggering immediately — it reports back whether the
  agent accepted the pull, so it cannot be deferred.
- The webhook description in the UI mentions the coalescing, since deployments now start a
  few seconds after the push.

## **Additional context**

Why the burst is coalesced rather than deduplicated on the first delivery: `buildx` pushes
the arch-specific child manifests before the tagged manifest list. Pulling on the first
delivery can therefore see the old tag ("No change") and would skip the delivery that
actually carries the update. Waiting for the burst to settle means the single pull sees the
final image.

The debounce window and cap are constants, matching the agent's `healthWatcher` debounce.
If a registry spreads its deliveries wider than the 30s cap, the extra pull reports "No
change" as before, and image polling still catches anything missed.

### DIFF
--- a/backend/internal/hub/applications/image_pull_debounce.go
+++ b/backend/internal/hub/applications/image_pull_debounce.go
@@ -13,16 +13,13 @@ import (
 )
 
 const (
-	// imagePullDebounceWindow is the quiet period a burst of image update
-	// triggers has to settle for before the pull is dispatched. Registries fan a
-	// single push out into several deliveries — GHCR publishes one package event
-	// per manifest of a multi-arch image — so pulling on the first delivery both
-	// floods history and notifications and can race the tag update itself.
+	// imagePullDebounceWindow is the quiet period a burst of image update triggers
+	// has to settle for before the pull is dispatched. One registry push can fan out
+	// into a delivery per manifest of a multi-arch image.
 	imagePullDebounceWindow = 5 * time.Second
 	// imagePullDebounceMaxDelay caps how long a continuous stream of triggers can
 	// postpone the pull.
 	imagePullDebounceMaxDelay = 30 * time.Second
-	imagePullDebounceTimeout  = 10 * time.Second
 )
 
 // DefaultImagePullDebouncer coalesces image update webhook deliveries. It is nil
@@ -37,8 +34,7 @@ type pendingImagePull struct {
 }
 
 // ImagePullDebouncer collapses bursts of image update triggers for the same
-// application into a single pull, so one registry push results in one history
-// entry and one notification instead of one per webhook delivery.
+// application into one pull, so one push yields one history entry and one notification.
 type ImagePullDebouncer struct {
 	mu       sync.Mutex
 	wg       sync.WaitGroup // tracks in-flight dispatches
@@ -69,8 +65,7 @@ func ScheduleImagePull(app *models.Application, source models.ApplicationEventSo
 }
 
 // Schedule queues a pull for appID once no further trigger arrives within the
-// debounce window. Repeated triggers restart that window, but never postpone the
-// pull beyond maxDelay from the first trigger of the burst.
+// debounce window, but never later than maxDelay after the first trigger.
 func (d *ImagePullDebouncer) Schedule(appID string, source models.ApplicationEventSource) {
 	if d == nil {
 		return
@@ -83,12 +78,9 @@ func (d *ImagePullDebouncer) Schedule(appID string, source models.ApplicationEve
 
 	now := time.Now()
 	deadline := now.Add(d.maxDelay)
-	if d.maxDelay <= 0 {
-		deadline = now.Add(d.window)
-	}
 	if pending, ok := d.pending[appID]; ok {
 		if !now.Before(pending.deadline) {
-			// The hard cap elapsed, so the queued pull is already on its way out.
+			// Hard cap elapsed; the queued pull is already on its way out.
 			return
 		}
 		deadline = pending.deadline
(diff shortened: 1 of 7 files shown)


backend/internal/hub/applications/image_pull_debounce.go
```diff
@@ -0,0 +1,164 @@
...
+	if DefaultImagePullDebouncer == nil {
+		TriggerImagePull(app, source)
+		return
+	}
+	DefaultImagePullDebouncer.Schedule(app.Id, source)
+}
+
+// Schedule queues a pull for appID once no further trigger arrives within the
+// debounce window. Repeated triggers restart that window, but never postpone the
+// pull beyond maxDelay from the first trigger of the burst.
```
timokoessler: Why don't we trigger the pull right now and then just ignore all other events in the next 5 seconds for this application coming via webhook? I guess we will have the same issue for the repo webhook (content). Not sure if there is a way to create a generic debounce handler that just stores a identifier in memory for the interval to skip new triggers?

alex289: Because "trigger now" can pull the *old* image and then swallow the event that carries the new one.

A multi-arch `buildx --push` uploads the per-platform child manifests first (untagged, by
digest) and the tagged manifest list last — the index can only be pushed once the manifests
it references exist. GHCR creates a package version per pushed manifest and fires a
`package published` event for each, so the first delivery we see can be a child manifest,
while the tag still resolves to the previous image. On top of that, GitHub does not
guarantee delivery order or timing, so "first delivery" is not necessarily "first event".

alex289: AI answer on the second part 😅

```
Agreed on the shared mechanism, but I'd flip the semantics: "store an id and skip new
triggers for the interval" is the leading-edge variant, and for repo webhooks it is worse
than for image webhooks.

Image webhook deliveries in a burst are all the *same* intent ("pull whatever the tag points
at now"), so dropping some is only a question of which one wins. Repo webhook deliveries are
not interchangeable — each carries its own commit and we deploy exactly that SHA
(`StaticCommit(pushDetails.Commit, "")`). Skipping the later trigger means deploying the
older commit and silently discarding the newer one:


t+0s  push abc123 → sync starts
t+3s  push def456 → skipped, "already handled" → def456 never deploys


That's the opposite of what we want, and it's not hypothetical: today a push arriving while
a sync is in flight is dropped by `tryLockRepositorySync` (204, no history entry), and since
`pollRepositories` only looks at `sync_type = polling`, nothing catches it up — the commit
waits for the next push or a manual sync. An interval-based skip would widen that window
from "while syncing" to "for the whole interval".

Also worth noting the fan-out itself doesn't really exist for git: one `git push` is one
push event per ref. What we see there is provider retries (already covered by the in-flight
lock) and genuinely distinct pushes in quick succession.

So if we extract something shared, I'd generalize what `ImagePullDebouncer` already does —
coalesce per key over a quiet window with a max-delay cap, **last trigger wins** — rather
than "first wins, rest dropped". For repos the key would have to be repo + branch (a push to
two branches must deploy both) and the payload the newest commit, and it needs to compose
with `TryLockRepositorySync` instead of replacing it.

That's a chunk of design on its own and the repo side has no reported noise problem, so I'd
keep it out of this PR. Happy to open a follow-up issue for the extraction plus the
"newer commit dropped while syncing" gap, which I think is the more valuable half.
```

backend/internal/hub/applications/image_pull_debounce.go
```diff
@@ -0,0 +1,164 @@
...
+	// imagePullDebounceWindow is the quiet period a burst of image update
+	// triggers has to settle for before the pull is dispatched. Registries fan a
+	// single push out into several deliveries — GHCR publishes one package event
+	// per manifest of a multi-arch image — so pulling on the first delivery both
+	// floods history and notifications and can race the tag update itself.
+	imagePullDebounceWindow = 5 * time.Second
+	// imagePullDebounceMaxDelay caps how long a continuous stream of triggers can
+	// postpone the pull.
+	imagePullDebounceMaxDelay = 30 * time.Second
+	imagePullDebounceTimeout  = 10 * time.Second
```
timokoessler: This timeout is only used once for a db query, a bit confusing maybe? But please do not add another long comment :D

backend/internal/hub/applications/image_pull_debounce.go
timokoessler: It's also possible to explain this in one or two sentences :D

backend/internal/hub/applications/image_pull_debounce.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	if d.stopped {
+		return
+	}
+
+	now := time.Now()
+	deadline := now.Add(d.maxDelay)
+	if d.maxDelay <= 0 {
```
timokoessler: It's a constant that is never below 0?

timokoessler (approved):

timokoessler (approved):

